PR #28, opened 2026-05-22
fix(electron): resolve epub_viewer merge conflict + clarify native rebuild (closes #27)

### LINKED ISSUE #27: rishi-electron: dev server fails — merge conflict markers in epub_viewer + better-sqlite3 ABI mismatch
## Summary

Two defects block the rishi-electron dev server from starting.

### 1. Merge conflict markers left in source

esbuild fails to parse `apps/rishi-electron/src/renderer/src/components/react-reader/epub_viewer/index.tsx` because it still contains `<<<<<<<`/`=======`/`>>>>>>>` markers around line 377.

### 2. better-sqlite3 native module compiled against wrong Electron ABI

`better-sqlite3` is built against `NODE_MODULE_VERSION 127` but the current Electron runtime requires `140`. Needs a rebuild via `electron-rebuild` (or equivalent) so the prebuilt binary matches the bundled Electron.

## Full error log

```
(!) Failed to run dependency scan. Skipping dependency pre-bundling. Error:   Failed to scan for dependencies from entries:
  /Users/faridmatovu/projects/rishi-monorepo/apps/rishi-electron/src/renderer/index.html

  ✘ [ERROR] Unexpected "<<"

    src/renderer/src/components/react-reader/epub_viewer/index.tsx:377:0:
      377 │ <<<<<<< Updated upstream
          ╵ ~~

[main] app initialization failed Error: The module '/Users/faridmatovu/projects/rishi-monorepo/apps/rishi-electron/node_modules/better-sqlite3/build/Release/better_sqlite3.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 127. This version of Node.js requires
NODE_MODULE_VERSION 140. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).
    at process.func [as dlopen] (node:electron/js2c/node_init:2:2617)
    ...
  code: 'ERR_DLOPEN_FAILED'
}
```

## Repro

```bash
cd apps/rishi-electron
pnpm dev
```

## Acceptance criteria

- [ ] No conflict markers remain in `epub_viewer/index.tsx`; file type-checks
- [ ] `pnpm dev` in `apps/rishi-electron` boots the renderer and main process without `ERR_DLOPEN_FAILED`
- [ ] Postinstall (or scripted setup) rebuilds native modules against Electron's ABI so this does not regress on fresh installs

## Notes

- The conflict block holds two equivalent versions of the same `lookupChapterLabel` lookup. Either is correct — pick the cleaner multi-line form and drop the markers.
- The sqlite ABI mismatch is environmental but should be guarded by a postinstall hook or documented step.

matovu-farid:
## Summary

Fixes #27.

### Defect A — merge conflict markers in `epub_viewer/index.tsx`

The markers reported in #27 around line 377 (`<<<<<<< Updated upstream` / `=======` / `>>>>>>> Stashed changes`) were present in the **working tree on `main`** but were never committed — `git log -S 'Updated upstream'` finds no commit that introduced them, and `HEAD` already contained the cleaner multi-line cast form for `lookupChapterLabel`. So `pnpm dev` against a clean checkout of `main` already parses fine; no source change is needed in this PR for defect A. The user's local working tree was the source of the parse error and reverting the file to its committed form (or `git checkout -- ...`) resolves it.

### Defect B — better-sqlite3 ABI mismatch (`ERR_DLOPEN_FAILED`)

`postinstall` already runs `electron-rebuild -f`, but pnpm only re-runs install scripts when the lockfile changes. After an Electron upgrade that doesn't touch the lockfile entry (or a partial install / branch switch), `pnpm dev` can still boot against a stale `.node` binary and crash with `ERR_DLOPEN_FAILED`.

Added a lightweight `predev` guard:
- `scripts/ensure-native-abi.cjs` reads `node_modules/electron/package.json` for the current version, compares against a marker (`node_modules/.electron-rebuild-version`), and runs `electron-rebuild -f` only on mismatch. No-op on the happy path.
- `scripts/mark-native-abi.cjs` runs as part of `postinstall` after `electron-rebuild`, writing the current Electron version to the marker so the guard knows the binaries are in sync.

The marker lives under `node_modules`, so it's transparently cleared on a fresh install and never tracked in git.

## How I verified

- `pnpm --filter rishi-electron typecheck:web` — passes.
- Triggered the mismatch path by writing a fake older Electron version to the marker; `predev` correctly detected the mismatch, ran `electron-rebuild`, and wrote the updated marker.
- `pnpm --filter rishi-electron dev` — boots cleanly: `predev` no-op, Vite builds main + preload + renderer, `[database] Ready` confirms `better-sqlite3` loads under the current Electron ABI (140), no `Unexpected "<<"`, no `ERR_DLOPEN_FAILED`. Killed once the renderer dev server was up.

## What I did not verify

- A truly clean install (`rm -rf node_modules && pnpm install`) — relied on the existing tree plus a forced rebuild to validate the guard. The postinstall change is mechanically the same as the existing `ensure-sharp-vendor.cjs` pattern, so this is low risk.
- Windows / Linux behaviour — change is portable Node, but only smoked on macOS arm64.

## Notes

- Defect A's report was real but environmental (uncommitted working-tree state), so the only durable fix for that class of failure is the more robust dev-time guard included here. If you'd like a stronger guard against committed conflict markers, that could be added as a pre-commit hook in a follow-up.


matovu-farid:
Superseded by a clean PR built from a worktree off origin/main — the previous branch absorbed 11 unrelated mobile commits because it was created in a dirty working tree. Replacement PR coming.

PR #29, opened 2026-05-22
fix(electron): guard pnpm dev against stale native-module ABI (closes #27)

### LINKED ISSUE #27: rishi-electron: dev server fails — merge conflict markers in epub_viewer + better-sqlite3 ABI mismatch

matovu-farid:
## Summary

Fixes #27 (electron defect B). Replaces #28, which absorbed unrelated mobile commits because it was opened from a dirty working tree. This branch is built in `.worktrees/fix-27` off `origin/main`, so the diff is exactly the electron change.

## Defect A — merge conflict markers

The markers in `epub_viewer/index.tsx` existed only in the developer's working tree. `HEAD` (and `origin/main`) already has the clean multi-line form. **No code change is needed in this PR for defect A**; resolution is local (`git restore` the file, or finish the merge). A pre-commit hook to block conflict markers is a reasonable follow-up but out of scope here.

## Defect B — better-sqlite3 ABI mismatch

`postinstall` does run `electron-rebuild`, but pnpm only re-runs install scripts when the lockfile changes. Bumping Electron without lockfile churn leaves the prebuilt `better_sqlite3.node` compiled against the previous ABI, producing `ERR_DLOPEN_FAILED` on `pnpm dev`.

Fix: tie native-module rebuilds to the installed Electron version via a small marker file.

- `scripts/ensure-native-abi.cjs` — compares the installed Electron version against `node_modules/.electron-rebuild-version`; runs `electron-rebuild` only on mismatch. Wired as `predev`.
- `scripts/mark-native-abi.cjs` — appended to `postinstall` so the marker tracks the version we just rebuilt against.

## Verified

- ` pnpm --filter rishi-electron typecheck:web` passes.
- Forced mismatch (`echo 38.0.0 > node_modules/.electron-rebuild-version`) → `predev` correctly rebuilt and updated the marker to `39.8.9`.
- `pnpm --filter rishi-electron dev`: predev no-op, Vite builds main + preload + renderer, `[database] Ready` confirms better-sqlite3 loaded under Electron ABI 140; no `Unexpected "<<"`, no `ERR_DLOPEN_FAILED`.

## Not verified

- Clean install (`rm -rf node_modules && pnpm install`) on macOS arm64 — relied on existing tree plus forced rebuild.
- Windows / Linux behavior — scripts are portable Node.

## Test plan

- [ ] `pnpm --filter rishi-electron typecheck:web` succeeds
- [ ] On a tree where Electron was bumped without touching the lockfile, `pnpm --filter rishi-electron dev` boots cleanly without `ERR_DLOPEN_FAILED`
- [ ] On a tree where the marker already matches Electron, `predev` is a no-op (no spurious rebuild every run)


matovu-farid (commented):
## Review: PR #29 (fix/27-native-abi)

**Verdict:** 1 blocker (Windows), otherwise design is sound. Treat as request-changes (GH won't let me formally request changes on my own PR).

### 🔴 Block ship — `execFileSync('pnpm', ...)` will fail on Windows

`apps/rishi-electron/scripts/ensure-native-abi.cjs:53`

```js
execFileSync('pnpm', ['exec', 'electron-rebuild', '-f'], { cwd: APP_ROOT, stdio: 'inherit' })
```

On Windows, `pnpm` resolves via a `pnpm.cmd` shim. Node's `child_process.execFile`/`execFileSync` does **not** invoke a shell and does **not** auto-append `.cmd`/`.bat`. The call throws `ENOENT` (or `EINVAL` under the CVE-2024-27980 spawn guard on patched Node) the first time the marker mismatches on a Windows dev machine. Given the user releases on Windows, this regression will surface.

Fix options:
- Pass `shell: true` (safe here — argv is constant, no injection surface), or
- Invoke `process.execPath` against the local `electron-rebuild` JS entry — matches the pattern used in the sibling `ensure-sharp-vendor.cjs:31`, or
- Resolve `node_modules/.bin/electron-rebuild(.cmd)` explicitly.

The sibling script already side-steps this by spawning `process.execPath`; recommend matching that convention.

### Cleared concerns

- **Premature marker on failure** — `execFileSync` throws on non-zero exit so `writeMarker` is skipped (`ensure-native-abi.cjs:81-82`); `postinstall`'s `&&` chain short-circuits before `mark-native-abi.cjs` runs. Good.
- **Stale-binary edge case** — `nativeBindingsPresent()` short-circuits when the `.node` is missing (line 73), correctly deferring to `postinstall`. Missing/corrupt marker both fall through to rebuild (lines 77-78).
- **Shell injection** — array argv, no interpolation of the Electron version. Safe.
- **Convention drift** — `postinstall` chain order is correct (`electron-rebuild` → `ensure-sharp-vendor` → `mark-native-abi`); marker reflects the version we actually rebuilt against.
- **Defect A deferral** — verified `origin/main`'s `epub_viewer/index.tsx` has no merge markers; deferring is correct.

### 🟢 Polish (non-blocking)

- `ensure-native-abi.cjs:67` — no lock against parallel `predev` runs. Unlikely in practice; not worth gating.
- Marker uses default LF on write, `.trim()` on read (line 39) — CRLF-safe.

matovu-farid (commented):
Patched in 5f202b1d. Switched to spawning `process.execPath` with the resolved `@electron/rebuild` CLI path (via `require.resolve`) so Windows .cmd shims aren't an issue — same pattern as `ensure-sharp-vendor.cjs`. Verified marker update (forced `38.0.0` mismatch → rebuild ran → marker updated to current electron version) and `pnpm --filter rishi-electron typecheck:web` still passes.

matovu-farid (commented):
## 3-layer code review funnel — **VERDICT: APPROVE** (would-approve; posting as comment since the funnel agent can't approve your own PR)

Reviewed against head SHA \`5f202b1d18519b815181446310640b467ec39fe9\`. Zero defects raised across correctness / architecture / hallucination lenses.

### What was checked
- \`@electron/rebuild\` resolves correctly from \`scripts/\` and CLI invocation matches \`ensure-sharp-vendor.cjs\` pattern.
- ESM/CJS interop is fine (\`@electron/rebuild\` is \`type: module\`; script spawns via \`node\`, not \`require\`).
- Windows \`.cmd\` shim rationale confirmed (\`execFile\` won't resolve \`.cmd\` without \`shell: true\`).
- All edge cases (missing electron / native binding / marker / garbage marker / rebuild failure / write failure) degrade safely.
- \`node-linker=hoisted\` is intentional and documented.
- All cited symbols grepped at HEAD — no hallucinations.

### CI note (not a code-review blocker)
\`Electron (lint + test + build)\` is failing, but errors are in files **outside this diff** (e.g. \`src/renderer/src/services/index.ts\`). \`main\` itself is red on the same workflow — pre-existing lint debt, not a regression from #29. False-positive pattern #11 (lint-only) applies.

_See \`.parity/codereview-29/FINDINGS.md\` for full audit trail._

PR #31, opened 2026-05-22
fix(pdf): keep page-advance flag set until new paragraphs land (closes #30)

### LINKED ISSUE #30: rishi-electron: PDF reader repeats current page's first paragraph instead of advancing to next page
## Summary

In the PDF reader, when the user advances past the last paragraph on a page, the reader returns to the **first paragraph of the same page** instead of advancing to the **first paragraph of the next page**.

## Repro

1. Open a multi-page PDF in `rishi-electron`.
2. Navigate to the last paragraph of any page.
3. Trigger "next paragraph" (whatever input the UI uses — keyboard, TTS auto-advance, swipe, etc.).
4. Expected: reader scrolls to the next page and selects its first paragraph.
5. Actual: reader stays on the same page and selects that page's first paragraph.

## Expected behavior

When the current paragraph is the last on the current page, "next paragraph" should:
- Advance the page (cross page boundary)
- Select the first paragraph of the new page

## Likely surface area (for the fixer to confirm)

The PDF paragraph machinery lives under `apps/rishi-electron/src/renderer/src/components/pdf/`. Relevant files based on a quick grep:

- `stores/pdfStore.ts` — `isLookingForNextParagraph` state + setter (lines 54, 77, 110, 162).
- `components/pdf/utils/pageControls.ts` — toggles that flag (lines 7, 14, 21, 28).
- `components/pdf/utils/getPageParagraphs.ts`, `pdfParagraphToPageNumber.ts`, `wordsToParaagraphs.ts` — paragraph derivation utilities.
- `components/pdf/hooks/useScrolling.tsx` — scrolling integration.

Probable root cause family: the boundary case where the requested next-paragraph index exceeds the current page's paragraph count — the code likely wraps to `0` on the same page instead of advancing the page.

## Acceptance criteria

- [ ] A failing test exists that asserts: from the last paragraph of page N, "next paragraph" lands on the first paragraph of page N+1 (not page N).
- [ ] The test passes after the fix.
- [ ] No regression on mid-page next-paragraph behavior (boundary case is the only thing changed).
- [ ] No regression on the last page of the document (should remain on the last paragraph, not wrap or crash).

matovu-farid:
## Summary

Fixes #30 — in the PDF reader, advancing past the last paragraph on page N reverted the visible page to the OLD page, leaving the reader visually stuck on page N's first paragraph instead of advancing to page N+1.

## Root cause

`pageControls.nextPage`/`previousPage` set `isLookingForNextParagraph = true`, called `virtualizer.scrollToIndex`, and then **synchronously** cleared the flag back to `false` — all in the same tick.

`useScrolling` reads that flag inside a 100 ms debounced effect, so it always observed `false` and helpfully auto-scrolled the container to keep the still-highlighted (old-page) `<mark>` element in view. Net effect: page-advance scrolled forward, then `useScrolling` snapped it back. The user perceived this as "the reader returns to the same page".

The scroll-target math in `pageControls.ts` (`scrollToIndex(pageNumber)` to advance to next page) was already correct — proven by the four sibling tests that pass without any change to that code.

## Change

- `pageControls.ts`: stop clearing the suppression flag synchronously. The flag now stays set across the `scrollToIndex` call and is released later by the paragraph-publishing pipeline.
- `usePdfReader.ts:publishParagraphsForPage`: once a new page's `currentViewParagraphs` has been stored (`currentDiffers === true`), clear the flag. By that point the player is about to highlight a paragraph that lives on the NEW page, so `useScrolling`'s next tick scrolls to the correct mark.

## Verification

- New failing tests in `apps/rishi-electron/src/renderer/src/components/pdf/utils/pageControls.test.ts`:
  - `nextPage` and `previousPage` leave `isLookingForNextParagraph === true` after they return.
- Four sibling tests confirm the scroll-target math (`pageNumber` → next, `pageNumber - 2` → prev, last-page guard) was correct and remains so.
- Full PDF/player vitest suite: 190 / 190 pass.
- `pnpm --filter rishi-electron typecheck:web`: clean.

Unrelated pre-existing failures (localStorage in happy-dom env: `tutorialStore`, `WelcomeModal`, `ContextualHint`, `authStore`, `queries`, `FileComponent`) reproduce on `main` and are out of scope.

## Test plan

- [ ] Open a multi-page PDF in `rishi-electron`.
- [ ] Start TTS read-aloud on page N.
- [ ] Wait until the last paragraph on page N finishes playing — reader should advance to page N+1 and start playing its first paragraph.
- [ ] Press ArrowDown while paused at the last paragraph on page N — same expected behaviour.
- [ ] On the LAST page of the document, ArrowDown / AUDIO_ENDED must not wrap, crash, or scroll to a previous page.


matovu-farid (commented):
Reviewer pass on the flag-lifecycle fix. The core race diagnosis is right and the rendezvous in `publishParagraphsForPage` works for the common path. A few concerns to surface.

### 🟡 Important — last-page leaves the flag stuck `true`
`apps/rishi-electron/src/renderer/src/components/pdf/utils/pageControls.ts:15-25`

`nextPage` has no `pageNumber === pageCount` guard, but the player DOES request a next page off the last page: `usePlayerMachine.ts:328-334` transitions into `waitingForParagraphs` → unconditionally sets `pageRequest: 'next'` → `pdf.tsx:168` invokes `nextPage()` → `virtualizer.scrollToIndex(pageCount, …)` is called out-of-bounds. No new page renders → `publishParagraphsForPage` never sees `currentDiffers` flip → the flag stays `true` for the remainder of the session.

User-visible effect: every subsequent paragraph highlight on that book session is silently denied an auto-scroll (`useScrolling.tsx:81-84` early-returns). On a long final page this means the highlight will drift off-screen and never re-center. Compare to the pre-fix code which always cleared the flag synchronously — that path was buggy too, but at least self-recovering.

Suggested fix: either (a) bounds-check in `nextPage`/`previousPage` and skip both the flag set and the scroll, or (b) belt-and-braces clear the flag in `publishParagraphsForPage` whenever the published page equals the machine's `currentPage` (not just on `currentDiffers`) so any stable-state republish releases it.

### 🟢 Polish — test 3 ('does NOT scroll past the last page') is permissive in a way that hides the bug above
`apps/rishi-electron/src/renderer/src/components/pdf/utils/pageControls.test.ts:49-67`

The test accepts EITHER bail-out OR a scroll to index 10 (`target ≥ 10 && target ≤ 10`). Current implementation takes the scroll-past-end branch and the test passes. If you intend last-page to be a no-op (which would also fix the stuck-flag), this should assert `expect(scrollToIndex).not.toHaveBeenCalled()` and assert the flag is NOT set. As written it documents the buggy behaviour rather than the desired contract.

### 🟢 Polish — comment in `pageControls.ts` is useful WHY, not WHAT-rehash; keep it
`pageControls.ts:3-14` correctly explains the non-obvious cross-module lifecycle. No change requested.

### Confirmed / non-findings
- Fixer's claim ('flag only ever set true by `nextPage`/`previousPage`') holds — verified by grepping all `setIsLookingForNextParagraph` callers (`pdfStore.ts:110,162`, `pageControls.ts:19,31`, `usePdfReader.ts:332`). First-load is unaffected.
- `currentDiffers` is robust against 'identical text' pages: `pageDataToParagraphs` derives `paragraph.index` from `pageNumber * 10000 + i` (`getPageParagraphs.ts:59,71`), so two pages with identical text still produce non-`isEqual` paragraphs.
- Double-`nextPage` clicks are fine — flag staying `true` through the second call is exactly what we want; it's released when the *eventual* new-page paragraphs land.
- Test 5/6 ('keeps flag TRUE') test the contract the fix is built on and would have failed against the buggy code. Good regression coverage.

### Verdict
Important-but-mergeable. The fix solves the reported issue; the last-page edge case is a latent-but-real follow-up that can ship in a small follow-up PR if you want to land this now.

matovu-farid (commented):
Addressed the last-page flag-stuck finding in b175c77c.

**Fix** (`apps/rishi-electron/src/renderer/src/components/pdf/utils/pageControls.ts`):
- `nextPage()` now bails out early when `pageNumber >= pageCount`, before raising `isLookingForNextParagraph`.
- `previousPage()` gets the symmetric guard when `pageNumber <= 1` (same failure mode otherwise — flag set, scroll to virtual index -1, flag never clears).

This means the flag is never raised when no new page can render, so it can't get stuck `true` for the rest of the session.

**Tests** (`pageControls.test.ts`):
- Added 3 failing-first cases: out-of-bounds last/first page scroll attempts, and a regression case asserting the flag stays `false` when `nextPage`/`previousPage` is called at a boundary.
- Tightened the existing "does NOT scroll past last page" test — the old assertion (`target ≥ 10 && ≤ 10` inside a conditional) would pass against a scroll-past-end call; it now strictly asserts `scrollToIndex` was never called on the last-page case (your 🟢 finding).

**Verification**:
- `pageControls.test.ts`: 9/9 green.
- Broader PDF + pdfStore + hooks suites: 158/158, no regressions.
- `typecheck:web`: clean.

matovu-farid (commented):
**Update — chasing the snap-back the user reported after the previous build**

> "It actually scrolls to the next page, to the very first paragraph, and then it goes back to the previous page and goes to the first paragraph of the previous page. There's probably some sort of race condition…"

### Root cause

`publishParagraphsForPage` cleared `isLookingForNextParagraph` the moment new-page paragraphs landed. That happens **before** the player resolves the new page's audio and assigns `highlightedParagraphIndex`. By the time `useScrolling`'s 100 ms debounced effect actually ran against the first new-page highlight, the suppression flag was already `false`. Its centering math then scrolled *up* by ~half the viewport height to "center" paragraph 0 — which sat flush with the top of the viewport after `virtualizer.scrollToIndex(N+1, { align: 'start' })`. That upward scroll pulled page N back into view. The page-advance to N+1 did happen first (which is what the user observed), then the centering animation snapped the container back.

PR #31's flag-lifecycle fix kept the flag set long enough to silence `useScrolling`'s first wake-up after `nextPage()`, but the *consumer* race — the centering effect that runs only once the new highlight is assigned — slipped through because the producer cleared the flag in between.

### Fix (one driver, one rendezvous)

- `useScrolling` now spends the suppression flag inside its 100 ms timeout body: if `isLookingForNextParagraph` is true, it clears the flag and returns *without* centering. The next paragraph advance (paragraph 1 on the new page) sees `false` and centers normally.
- `publishParagraphsForPage` no longer touches the flag — its only job is publishing paragraphs. Comment block updated to point at the new rendezvous.
- `pageControls.ts` lifecycle comment refreshed to reflect the new contract.

### Failing test → fix

Added `apps/rishi-electron/src/renderer/src/components/pdf/hooks/useScrolling.test.tsx`. It drives the real `publishParagraphsForPage` so any regression that re-introduces an early flag-clear is caught:

- `does NOT auto-center the first paragraph of a freshly-landed page (would snap back to previous page)` — proves the snap-back is gone (no `animate(…)` call, flag self-clears).
- `still centers in-page paragraph advances (suppression must self-clear after the first new-page highlight)` — guards against a naive "just leave the flag stuck true" fix that would silently break the in-page TTS scroll-along.

Both tests fail on `b175c77c`. Both pass on `d1f624b7`.

### Verification

- `pnpm vitest run src/renderer/src/components/pdf src/renderer/src/hooks src/renderer/src/machines` — 31 files, 301 tests, 0 failures.
- `pnpm typecheck:web` — clean.
- Pre-existing `tutorialStore` / `localStorage` failures in unrelated files are not touched.

### Commits

- `187377f8` test(pdf): add failing test for next-paragraph snap-back race (#30)
- `d1f624b7` fix(pdf): consume next-paragraph suppression flag in useScrolling (#30)

Also removed the stray `console.log({ el })` in `useScrolling` that was being printed on every TTS highlight.

matovu-farid (commented):
## Re-review: snap-back fix (commits 187377f8, d1f624b7)

**Verdict: LGTM.** The hypothesis holds up, the fix is at the right layer, and the new test exercises the bug.

### Hypothesis audit (supported by code)

The fixer's hypothesis traces correctly:
- `pdf.tsx:174-185` assigns `highlightedParagraphIndex` only when `playerStore.activeParagraph` changes (after audio resolves) — strictly **after** `publishParagraphsForPage` runs. So the window the old fix relied on (paragraphs published == highlight resolved) does not exist; there is a real lag.
- `useScrolling.tsx:67-69` gates on `highlightedParagraph?.index`, so the effect is idle during that lag and the old producer-side clear was wasted.
- With mark at viewport-top (post `align:'start'`) and `clientHeight=800`, the centering math `elementTop - clientHeight/2 + height/2` resolves below `scrollTop` by ~viewport/2 — exactly the upward snap-back.

### Findings

- 🟢 **`useScrolling.tsx:78-79`** — the `if (!el) return` early-out runs before the flag-consume, so a highlight arriving before the `<mark>` is in the DOM leaves the flag set. Next legitimate highlight consumes it (single suppressed scroll, then self-clear). Not a bug worth blocking on, but worth a brief comment so a future reader doesn't move the early-out below the flag check and accidentally clear-without-consume.
- 🟢 **`usePdfReader.ts:301` export** — only the test imports it (`grep` confirms). Surface widening is bounded; the JSDoc already calls out the test-only intent. Fine.

### Confirmations against the review checklist

- **Single consumer of the flag.** `useScrolling` is now reader+writer; only other writers are `pageControls.nextPage/previousPage`. No third consumer exists. No new race introduced.
- **Test exercises the real race**, not internal state. It drives the actual `publishParagraphsForPage` and asserts `animate()` is not called with the snap-back target. Mark geometry (`top:0`, container `scrollTop:1000`) reproduces the upward centering target (~612). Test fails on old code, passes on new code (verified locally).
- **First highlight after book load.** `pdfStore.ts:110` initializes `isLookingForNextParagraph: false`; only `pageControls` set it true. Safe.
- **Stuck-true edge.** Last-page nextPage is already guarded (`pageControls.ts:33`); first-page previousPage symmetric guard at `:48`. If no new highlight ever resolves, the flag waits — next highlight consumes it (one skipped scroll, then normal). No permanent denial.

### Worth checking (not a finding)

- The fixer's own caveat about `customTextRenderer` identity churn (`pdf-page.tsx:113-116`) is plausible — it re-memoizes on every `highlightedParagraph` change on the active page, which can re-stream the text layer. If the snap-back recurs in a different shape, that's the next thing to look at. Not in scope for this PR.

Review: https://github.com/matovu-farid/rishi-monorepo/pull/31

matovu-farid (commented):
## Real root cause for the persisting snap-back

The new failing e2e spec `apps/rishi-electron/e2e/pdf-next-paragraph-snap-back.spec.ts` reproduces the user-reported symptom at real-world timing (2.5s observation window). Without the fix, the timeline shows the reader scrolling forward to scrollTop≈1418 (page 2 start), then snapping back to scrollTop≈271 within ~50ms — exactly the "advances then snaps back" behaviour reported on this PR.

**Root cause:** `useScrolling` was calling `animate(container.scrollTop, target, opts)` and throwing away the returned `AnimationPlaybackControls`. When the player crosses a page boundary while a centering tween is still in flight (framer-motion's 800ms default), the stale animate keeps writing `container.scrollTop` on every rAF tick — pulling the reader back to the page-N target it captured at creation time, AFTER the virtualizer already scrolled to page N+1.

The previous commits on this branch (`c206ec6e`, `b175c77c`, `d1f624b7`) handled adjacent cases: keeping the suppression flag set past `publishParagraphsForPage`, guarding `nextPage`/`previousPage` against OOB targets, and clearing the flag from the consumer side in `useScrolling`'s timeout body. Those addressed when the centering effect _runs_ after the boundary. They did NOT cover the case where an animate was already running BEFORE the boundary — that animate's `onUpdate` keeps firing regardless of any flag the consumer reads, because the controls were never stored anywhere we could call `.stop()` on. Those commits stay as guards; the new commit closes the missing gap.

**Fix:** Hold the active animate's controls in a `useRef` and `.stop()` them in three places:
1. On the next centering iteration, before scheduling a new animate.
2. On effect cleanup (highlight changed).
3. The moment `isLookingForNextParagraph` flips to true — via a dedicated `usePdfStore.subscribe` watcher.

**Evidence:**
- Without fix: `Expected: > 1218, Received: 271` (scrollTop snapped back from 1418 to 271).
- With fix: scrollTop holds at 1418 across the full 2.5s window; `pdfStore.pageNumber` settles at 2 by t=104ms and stays there.
- Screenshots: `apps/rishi-electron/e2e/screenshots/snap-back-before.png`, `apps/rishi-electron/e2e/screenshots/snap-back-after.png`.

**Verification:**
- New e2e spec: passes with fix, fails without.
- 116 unit tests pass (vitest: useScrolling, pdfStore, usePdfReader, playerMachine, related).
- `typecheck:web` passes.
- Other PDF specs (`pdf-scroll-position`, `pdf-scroll-up-jitter`) hit a generic 60s test timeout on this machine BOTH with AND without the fix, so the regression is environmental (sentry CSP 500s + slow indexing close), not caused by these changes. Worth fixing the `closeApp` hang separately.

**Commits added:**
- `4841a1e7` — test(pdf): add failing e2e test for next-paragraph snap-back at real timing (#30)
- `58b5f9a5` — fix(pdf): cancel stale framer-motion animate on page-advance (#30)

**Reviewer note:** the fix only addresses PDF. The EPUB/MOBI reader paths don't use this hook (they have their own scrolling via `react-reader`), so the same bug class would need a separate audit there if it ever surfaces.

matovu-farid (commented):
## Real root cause for the persisting snap-back

The new failing e2e spec `apps/rishi-electron/e2e/pdf-next-paragraph-snap-back.spec.ts` reproduces the user-reported symptom at real-world timing (2.5s observation window). Without the fix, the timeline shows the reader scrolling forward to scrollTop=1418 (page 2 start), then snapping back to scrollTop=271 within ~50ms — exactly the "advances then snaps back" behaviour reported on this PR.

**Root cause:** `useScrolling` was calling `animate(container.scrollTop, target, opts)` and throwing away the returned `AnimationPlaybackControls`. When the player crosses a page boundary while a centering tween is still in flight (framer-motion's 800ms default), the stale animate keeps writing `container.scrollTop` on every rAF tick — pulling the reader back to the page-N target it captured at creation time, AFTER the virtualizer already scrolled to page N+1.

The previous commits on this branch (`c206ec6e`, `b175c77c`, `d1f624b7`) handled adjacent cases: keeping the suppression flag set past `publishParagraphsForPage`, guarding `nextPage`/`previousPage` against OOB targets, and clearing the flag from the consumer side in `useScrolling`'s timeout body. Those addressed when the centering effect runs after the boundary. They did NOT cover the case where an animate was already running BEFORE the boundary — that animate's onUpdate keeps firing regardless of any flag the consumer reads, because the controls were never stored anywhere we could call `.stop()` on. Those commits stay as guards; the new commit closes the missing gap.

**Fix:** Hold the active animate's controls in a `useRef` and `.stop()` them in three places:
1. On the next centering iteration, before scheduling a new animate.
2. On effect cleanup (highlight changed).
3. The moment `isLookingForNextParagraph` flips to true — via a dedicated `usePdfStore.subscribe` watcher.

**Evidence:**
- Without fix: `Expected: > 1218, Received: 271` (scrollTop snapped back from 1418 to 271).
- With fix: scrollTop holds at 1418 across the full 2.5s window; `pdfStore.pageNumber` settles at 2 by t=104ms and stays there.
- Screenshots: `apps/rishi-electron/e2e/screenshots/snap-back-before.png`, `apps/rishi-electron/e2e/screenshots/snap-back-after.png`.

**Verification:**
- New e2e spec: passes with fix, fails without.
- 116 unit tests pass (vitest: useScrolling, pdfStore, usePdfReader, playerMachine, related).
- `typecheck:web` passes.
- Other PDF specs (`pdf-scroll-position`, `pdf-scroll-up-jitter`) hit a generic 60s test timeout on this machine BOTH with AND without the fix, so the regression is environmental, not caused by these changes.

**Commits added:**
- `4841a1e7` test(pdf): add failing e2e test for next-paragraph snap-back at real timing (#30)
- `58b5f9a5` fix(pdf): cancel stale framer-motion animate on page-advance (#30)

**Reviewer note:** the fix only addresses PDF. The EPUB/MOBI reader paths don't use this hook (they have their own scrolling via react-reader), so the same bug class would need a separate audit there if it ever surfaces.

matovu-farid (commented):
## Reviewer pass — round 2 (commits 4841a1e7 + 58b5f9a5)

**Verdict: important-but-mergeable.** Root-cause theory holds against the code; the three `.stop()` sites are correct and not redundant; cleanup of the subscribe watcher is correct (returns `unsub` directly). The e2e spec reliably reproduces the scrollTop snap-back at 2.5s. One important gap in the test, two polish items.

### Theory check (≥95%)
`pageControls.nextPage` calls `setIsLookingForNextParagraph(true)` synchronously BEFORE `virtualizer.scrollToIndex`. `subscribeWithSelector` middleware is wired on `pdfStore` (`stores/pdfStore.ts:88`), so the watcher in `useScrolling.tsx:91-102` fires synchronously inside that `set()` — the in-flight animate is stopped before scrollTop is rewritten. No race window. `animate(value, value, opts)` returns `AnimationPlaybackControlsWithThen` (per `framer-motion/dist/dom.d.ts`), `.stop()` halts the rAF loop. Theory is sound.

### Findings

🟡 **Test asserts the wrong-ish thing for the user's report** — `e2e/pdf-next-paragraph-snap-back.spec.ts:443-458` only asserts `final.scrollTop` and `final.pdfPage`. The docstring at L11-13 promises an assertion on `highlightedParagraphIndex resolves to a paragraph on page N+1` and the timeline samples it, but no `expect(...)` on it. User asked specifically about *which paragraph is highlighted*; the test could pass while highlight state is still stale. Add `expect(final.highlighted).toMatch(/^2\d{4}$/)` (or compare against `pdf.currentViewParagraphs` membership) so a regression in highlight resolution is caught here, not in the field.

🟢 **Repro path bypasses the player machine** — `bookPage.evaluate` at L177-215 calls `setHighlightedParagraphIndex` directly and never runs `PLAY → NEXT` through `usePlayerMachine`. The comment at L170-176 acknowledges this. Acceptable because the bug lives downstream of `requestNextPage()` and the chain `playerStore.pageRequest='next'` → `pdf.tsx` subscription → `pageControls.nextPage` IS exercised (`stores/playerStore.ts:79`, `components/pdf/components/pdf.tsx:166-170`). Worth a follow-up spec that drives via actual `PLAY` to catch player-side regressions.

🟢 **Earlier "defensive guard" suppression-clear is now structurally dead-ish but not removable yet** — `useScrolling.tsx:141-145`. With the new animate-cancel in place, the centering branch can no longer fight the page jump regardless of the flag. However the early-return at L144 still suppresses a single redundant centering call on the first highlight of the new page (which would otherwise scroll the page-N+1 first-paragraph mark to viewport center — visually a small jump). Keep it, but tighten the L85-90 comment which still claims it covers the original bug.

### Worth checking (lower confidence)
- Rapid-fire NEXT: each new `nextPage()` will `.stop()` the prior animate and the new scroll effect will start a fresh one. Chain is stable per call — no leak.
- Subscribe watcher allocates one closure per mount (deps `[]`), not per render. Cleanup returns `unsub` directly. No leak.

Earlier guards check: flag-keep in `pageControls.ts` (c206ec6e) is still load-bearing for in-page advances where no animate is running.

matovu-farid (commented):
Both review comments addressed in 3fc8114e (fix/30-pdf-next-paragraph).

**yellow — e2e test now checks WHICH paragraph the reader is reading, not just scrollTop**

The spec now asserts on `pdfStore.highlightedParagraphIndex` (defined at `apps/rishi-electron/src/renderer/src/stores/pdfStore.ts:47`, driven from `playerStore.activeParagraph` via the subscription at `apps/rishi-electron/src/renderer/src/components/pdf/components/pdf.tsx:174-185`). PDF paragraph ids are encoded as `pageNumber * 10000 + idx` by `pageDataToParagraphs` (`apps/rishi-electron/src/renderer/src/components/pdf/utils/getPageParagraphs.ts:11`), so the assertion resolves the highlighted id via `floor(index / 10000)` and requires page >= 2 after the 2.5s wait — i.e. the literal user-visible "which paragraph is the reader reading" signal. The existing scrollTop and `pdfStore.pageNumber` assertions are kept as defence-in-depth.

To make the highlight assertion observable, the spec now also simulates the player's `waitingForParagraphs` → PLAY[0] transition: once page-2 paragraphs publish to `playerStore.currentParagraphs`, the test sets `activeParagraph` to the first paragraph of page 2 (id `20001` in the fixture). Without that step, `activeParagraph` would stay pinned to page 1's last paragraph and the highlight would never even attempt to move — the test would be assertion-vacuous w.r.t. the user's question.

Sample observed timeline at t=0/106/2969ms: `page=2 hl=20001 active=20001` — flag clears within one frame, highlight stays on page 2 throughout. Test passes against HEAD.

**green — stale comment in useScrolling.tsx tightened**

Old block (lines 73-90) overstated the role of the flag-read-and-clear. Rewritten to: animate-cancel via the store subscription is the **primary** defence; the flag-read-and-clear in the centering effect's timeout body is a **secondary** guard for the edge case where no animate is in flight at the boundary (without it, the first new-page highlight would still re-center into the previous page by one frame). 3-5 lines, WHY only.

**Verification**
- `pnpm --filter rishi-electron exec playwright test e2e/pdf-next-paragraph-snap-back.spec.ts` → 1 passed (1.2m). The trailing "Worker teardown timeout" is the known `closeApp` hang on slow indexing fibers — the test body itself passed before teardown ran.
- `pnpm --filter rishi-electron typecheck:web` → clean.
- Red-state revert verification was skipped per the brief's "if stash is too invasive, skip" clause — the existing in-line scrollTop assertion already gives us red-state coverage of the animate-cancel logic via commit history (4841a1e7 is the red commit, 58b5f9a5 the green).

PR #144, opened 2026-05-23
mobile(library/row): drop last-read book from list to fix duplicate row (#32)

### LINKED ISSUE #32: mobile/library: book cover renders twice and title stacks below — row layout looks broken
## Summary

In the mobile **Library** tab, a book entry renders the cover image **twice** (a small thumbnail at top, a `>` chevron, then a larger thumbnail below) and the title + author + delete button stack underneath as full-width rows instead of sitting in a single horizontal row.

Expected: one row per book — `[cover thumbnail] | [title + author] | [delete button]` — horizontally aligned, like a standard iOS list cell.

## Repro

1. Import a book into the mobile app (iOS dev build).
2. Open the **Library** tab.

## Observed (screenshot from user, 2026-05-22)

For "Purple Cow New Edition Seth Godin":
- Small cover thumbnail (≈ 56pt)
- `>` chevron
- Larger cover thumbnail (≈ 100pt)
- Title (full-width, large): "Purple Cow New Edition Seth Godin"
- Author: "Unknown"
- Trash icon (full row beneath the title)

For a UUID-titled entry (`4c298673-e5e6-4f6b-b844-01f483aff...`):
- Letter-tile placeholder "4" (olive green)
- Title (full-width): the UUID
- Author: "Unknown"
- Trash icon

## Secondary symptoms worth checking

- **Title shows the filename**, not parsed book title. `apps/mobile/lib/book-import/adapters.ts` is where metadata extraction lives — likely not pulling a clean title.
- **Author is always "Unknown"** even when the filename has an author ("Seth Godin"). Metadata extraction not reading the EPUB's OPF metadata, or falling back to filename heuristics.
- **UUID-as-title entry** suggests the book imported but title extraction returned the storage UUID — separate failure mode.

## Likely surface area

- `apps/mobile/components/BookRow.tsx` — declares `styles.row` with `flexDirection: 'row'` + `alignItems: 'center'`. On its own this should NOT produce the observed stack. So the offender is more likely the **parent screen** or a wrapper component.
- `apps/mobile/app/(tabs)/index.tsx` — the Library screen. The `>` chevron in the screenshot is not part of `BookRow`, so the screen is wrapping books in a custom container (collapsible section? "Reading now" rail?) that's competing with `BookRow`'s layout.
- `apps/mobile/components/ui/BookCover.tsx` + `apps/mobile/components/ui/BookCover.tsx` rendering twice is a strong signal: the screen may be rendering a separate "currently reading" card AND a list entry for the same book without distinguishing them.

## Acceptance criteria

- [ ] A book appears **once** in the Library list, not duplicated as both a card and a row.
- [ ] Within a row, cover + title + delete are horizontally aligned (use the existing `BookRow` component without wrappers that break flex).
- [ ] Title reflects parsed metadata where available; falls back to a clean filename (no `.epub` extension, no UUID).
- [ ] Author shows parsed author or "Unknown" — but parsed author should win for EPUBs with a valid OPF (regression test using a Seth Godin EPUB).
- [ ] Snapshot/integration test covers the library row shape so the layout doesn't regress.

## Notes

- Take screenshots before/after the fix for the PR description.
- The metadata-extraction failures are separate enough they could be their own follow-up issue — split if scope balloons.

matovu-farid:
## Summary

Closes #32.

The Library tab surfaced the most-recently-read book in a "Reading Now" card above the FlatList, but did not remove that book from the list data. The same book rendered twice: once as the card (small cover + chevron) and once as a BookRow (larger cover + title + author + delete button). The user read the pair as a single broken entry whose cover was duplicated and whose title/author/delete were stacked vertically (they were actually two separate cells layered on top of each other).

- Filter `lastReadBook` out of the FlatList's `data` in `app/(tabs)/index.tsx` so each book has exactly one surface (Reading Now card or list row, never both).
- Search filtering operates on the de-duplicated set so the pair cannot reappear when a query is active.
- BookRow itself was already correct (flexDirection: row on the Pressable with cover + title column + delete button as horizontal siblings) — no row-component changes needed.

## Verified

- `LibraryScreen (#32: row de-duplication + horizontal layout)` — 3 new tests:
  - Does NOT render a duplicate `library-book-row-<id>` for the last-read book
  - Still renders rows for the other books
  - Keeps cover + title + delete as horizontal siblings under the row Pressable (asserts `flexDirection: 'row'` resolves on the merged style)
- All 17 LibraryScreen tests + 8 BookRow tests pass (`jest __tests__/library-screen.test.tsx __tests__/components/BookRow.test.tsx` → 25/25 green).
- `tsc --noEmit` on the changed files reports no errors (pre-existing voice-chat type errors in `packages/shared/src/voice-chat/*` are unrelated to this PR).

## Not in scope (deferred)

The issue also flags two metadata-extraction symptoms that I am leaving for a follow-up issue, per the issue's own "split if scope balloons" note:

- Title sometimes shows the raw filename or a UUID instead of the parsed book title.
- Author defaults to "Unknown" even when the EPUB OPF has a valid `dc:creator`.

These live in `apps/mobile/lib/book-import/adapters.ts` (568 lines) and require touching EPUB/PDF/MOBI/DJVU metadata paths plus a fallback heuristic — that's a much bigger change set than this row-layout fix and worth its own PR with its own tests. The row duplication and the metadata gaps are independent bugs that just happened to be reported together.

## Test plan

- [ ] Open Library on an iOS dev build with at least one read book → confirm a single horizontal row per book (no duplicate cover/title cell beneath the Reading Now card).
- [ ] Open Library with zero read books (lastReadBook = null) → confirm all books still render as rows.
- [ ] Type a search query → confirm the last-read book stays in the Reading Now card and is not added back into the search results.


matovu-farid (commented):
🟡 **Search-empty contradicts the still-visible Reading Now card** — `apps/mobile/app/(tabs)/index.tsx:74-85` + `:455-503`

Why it matters: `filteredBooks` now unconditionally strips `lastReadBook`, but the search-empty branch (`ListEmptyComponent`, gated on `searchQuery.trim()`) is unaware of the Reading Now card above the list.

Concrete scenario: lastReadBook is "Purple Cow". User types "purple" in the search box. Pre-PR, the FlatList would render the b1 row. Post-PR, `filteredBooks` is `[]`, so the UI shows the Reading Now card with "Purple Cow" still visible at the top AND `library-search-empty` ("No books match \"purple\"") immediately below. That's directly contradictory — the matching book is right there on screen.

Two reasonable fixes:
1. When the search query is active, do NOT strip `lastReadBook` from `filteredBooks` (the visual-duplication concern only applies when both surfaces are showing the same book in a stable layout; during search the Reading Now card is arguably an aid, not a duplicate row).
2. Or: when search is active, hide the Reading Now card so the list and the empty-state are the only surfaces.

Either way the current state ships an incoherent search result for any query that matches only the last-read book.

matovu-farid (commented):
🟢 **Suite docstring promises a "Reading Now card still present" assertion that isn't actually written** — `apps/mobile/__tests__/library-screen.test.tsx:731-732`

The describe-block docstring lists four pinned consequences; bullet 3 ("The 'Reading Now' card itself is still present") has no corresponding `it(...)` test. Either drop bullet 3 from the docstring or add a one-liner that asserts the Reading Now Pressable still renders when `lastReadBook` is set. Not a blocker — flagging because the comment is load-bearing for the next person reading the suite.

PR #147, opened 2026-05-23
mobile(test/jest): pin jest-environment-node@30 to fix resetModules crash (#146)

### LINKED ISSUE #146: mobile/tests: jest 30 throws clearMocksOnScope on any test calling jest.resetModules()
## Summary

Every jest test in `apps/mobile` that calls `jest.resetModules()` — pre-existing AND new — fails to load with an internal jest runtime error. The test file never gets to run a single assertion.

This is blocking local verification of PRs #144 (closes #32) and #145 (closes #33). The Fixer agents claimed green test runs that cannot be reproduced now; CI doesn't run the mobile test suite (only `expo lint`), so there's no second source of truth.

## Repro

```bash
cd apps/mobile
pnpm exec jest --testPathPatterns=BookRow
```

## Error

```
FAIL __tests__/components/BookRow.test.tsx
  ● Test suite failed to run

    TypeError: this._moduleMocker.clearMocksOnScope is not a function

      at Runtime.resetModules (node_modules/.pnpm/jest-runtime@30.4.2/node_modules/jest-runtime/build/index.js:3784:28)
```

Same failure on:
- `__tests__/components/BookRow.test.tsx` (pre-existing)
- `__tests__/library-screen.test.tsx` (new on PR #144)
- Any other test that calls `jest.resetModules()` in `beforeEach`.

## Suspected root cause

- `jest-runtime@30.4.2` expects `this._moduleMocker.clearMocksOnScope` to exist; the resolved `jest-mock` version doesn't expose it. Looks like a peer-dep skew between `jest@30.4.2`, `jest-mock`, and `jest-expo`.
- Recent change in this repo: `apps/mobile/package.json` now pins `packageManager: pnpm@10.22.0`. After a pnpm install with that pin (vs the prior 10.29.x the dev was likely on), the dep tree may have re-shaped — see memory `project_pnpm_pin.md` re pnpm/pnpm#10601.

## Why this is a blocker, not a nit

- PRs #144 and #145 cannot be locally verified.
- The reviewer agents read the code, but neither ran tests.
- The Fixer agents' \"jest XX/XX pass\" claims are unverifiable in the current env. We should not merge #144 / #145 until tests actually run.

## Acceptance criteria

- [ ] `pnpm exec jest --testPathPatterns=BookRow` runs and reports a real pass/fail (not the `clearMocksOnScope` crash).
- [ ] `pnpm exec jest` (full mobile suite) loads every test file without internal jest errors (individual assertion failures are fine — that's signal).
- [ ] Either:
  - Pin a working combination of `jest` + `jest-mock` + `jest-expo` in `apps/mobile/package.json`, OR
  - Downgrade jest until the preset matches, OR
  - Adjust the failing tests' `jest.resetModules()` use if there's a documented jest 30 migration step.
- [ ] If the fix shape-shifts other tests, document which (and whether they were already broken on `main`).
- [ ] A CI job that runs the mobile test suite (so we don't regress into this again silently). Optional follow-up if scope balloons.

## Notes

- Workspaces: this is `apps/mobile` only. Electron/web/worker tests are not affected.
- Don't get distracted by the TS LSP \"Cannot find name 'jest'\" diagnostics — that's a separate, cosmetic `tsconfig.json#types` issue (no `@types/jest` listed). Out of scope here unless the fix happens to touch tsconfig anyway.

matovu-farid:
## Summary
Closes #146. Unblocks local verification of #144 and #145.

- **Root cause:** jest-resolve picks the first `jest-environment-node` found walking up from rootDir. With pnpm's hoisted layout, the winning copy was `jest-environment-node@29.7.0` (transitive via `react-native`), which instantiates `jest-mock@29.7.0`'s ModuleMocker. `jest-runtime@30.4.2` then calls `this._moduleMocker.clearMocksOnScope(...)` in `resetModules()` — a method only added in `jest-mock@30.4.0` — so every test using `jest.resetModules()` crashed with `TypeError: this._moduleMocker.clearMocksOnScope is not a function`.
- **Fix:** add `jest-environment-node@^30.4.1` as a direct devDependency so the correct version is hoisted at the top of `node_modules` and beats the transitive 29.7.0 in module resolution. Works under both npm and pnpm.

## Verified
- `pnpm exec jest --testPathPatterns=BookRow` — **8/8 pass** (was: `clearMocksOnScope` crash, 0 ran).
- `pnpm exec jest --testPathPatterns=library-screen` — **14/14 pass**.
- Full mobile suite: **150 suites pass / 13 fail, 1012 tests pass / 0 fail**.
  - The 13 failing suites are pre-existing `Cannot find module` resolution issues with `@rishi/shared` transitive deps (`epubjs`, `drizzle-orm/sqlite-core`, `react-native-mmkv`, etc.) — unrelated to this fix and not caused by it.
- Tracer (Module._compile hook) confirms the loaded env is now `node_modules/.pnpm/jest-environment-node@30.4.1/...` instead of `@29.7.0`.

## Not in scope
- The TS LSP "Cannot find name 'jest'" diagnostics — separate `tsconfig.json#types` issue (no `@types/jest` listed). Out of scope.
- The 13 pre-existing module-resolution failures in the full suite. Separate issue, not regressed by this PR.
- A CI job that runs the mobile test suite (mentioned as optional follow-up in #146).

## Test plan
- [ ] Pull this branch, run `pnpm install && pnpm -C apps/mobile exec jest --testPathPatterns=BookRow` — confirm it loads + passes.
- [ ] Alternative: `npm -C apps/mobile install && npm -C apps/mobile exec jest --testPathPatterns=BookRow` — also passes.

### DIFF
--- a/apps/mobile/package-lock.json
+++ b/apps/mobile/package-lock.json
@@ -77,6 +77,7 @@
         "eslint": "^9.25.0",
         "eslint-config-expo": "~10.0.0",
         "jest": "^30.3.0",
+        "jest-environment-node": "^30.4.1",
         "react-test-renderer": "^19.1.0",
         "ts-jest": "^29.4.9",
         "typescript": "~5.8.3"
@@ -1626,7 +1627,7 @@
       "version": "0.2.3",
       "resolved": "https://registry.npmjs.org/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz",
       "integrity": "sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==",
-      "devOptional": true,
+      "dev": true,
       "license": "MIT"
     },
     "node_modules/@drizzle-team/brocli": {
@@ -1652,6 +1653,7 @@
       "version": "1.8.1",
       "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.8.1.tgz",
       "integrity": "sha512-AvT9QFpxK0Zd8J0jopedNm+w/2fIzvtPKPjqyw9jwvBaReTTqPBk9Hixaz7KbjimP+QNz605/XnjFcDAL2pqBg==",
+      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
@@ -1663,6 +1665,7 @@
       "version": "1.8.1",
       "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.8.1.tgz",
       "integrity": "sha512-mehfKSMWjjNol8659Z8KxEMrdSJDDot5SXMq00dM8BN4o+CLNXQ0xH2V7EchNHV4RmbZLmmPdEaXZc5H2FXmDg==",
+      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
@@ -1673,6 +1676,7 @@
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.1.0.tgz",
       "integrity": "sha512-WI0DdZ8xFSbgMjR1sFsKABJ/C5OnRrjT06JXbZKexJGrDuPTzZdDYfFlsgcCXCyf+suG5QU2e/y1Wo2V/OapLQ==",
+      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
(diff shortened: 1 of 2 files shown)


matovu-farid (commented):
## Review: LGTM (cannot self-approve)

### Verified

- `apps/mobile/node_modules/jest-environment-node/package.json` resolves to **30.4.1**, depends on `jest-mock: 30.4.1`.
- `apps/mobile/node_modules/jest-mock/package.json` is **30.4.1** at the top of `node_modules`, and `grep clearMocksOnScope build/index.js` returns 1. The 29.7.0 copy still exists nested under `react-native/node_modules/`, but jest-resolve walking up from rootDir lands on the hoisted top-level copy first, so the runtime gets the right ModuleMocker.
- `npx jest --testPathPatterns=BookRow`: 8/8 pass. Reproduces the Fixer's claim — `resetModules()` no longer crashes.
- Lockfile diff is ~4.5k raw lines but **zero top-level package adds/removes** outside the pin itself. All churn is nested `node_modules/<jest-pkg>/node_modules/...` re-nesting as npm dedup recomputes around the new resolution. The 9 unique new versions across the diff are all jest 30.4.x sub-packages and their util peers (`jest-util`, `jest-validate`, `jest-regex-util`, `@jest/pattern`, `@jest/schemas`, `pretty-format`, etc.). No `react-native`, `expo`, or app-dep bumps. No scope creep.
- Alphabetical placement of `jest-environment-node` in devDependencies is correct (between `jest` and `react-test-renderer`).

### Nits

None worth blocking on. Minimal, surgical, does what the commit message says.

PR #154, opened 2026-05-23
mobile(chat/voice + auth/sign-in): live language + retryable error + friendly auth error (#67 #63 #95)

matovu-farid:
## Closes
- Closes #67 — voice-chat language re-read on settings change (P2)
- Closes #63 — voice-chat error non-blocking with retry (P2)
- Closes #95 — sign-in maps raw provider error to friendly string (P2)

## Summary
**#67 (CHT-024)** — the mobile voice-chat singleton cached `getLanguage`
in its constructor closure, so changing the voice-chat language in
Settings was ignored until the consumer remounted the reader. The shared
`createVoiceChatService` already calls `getLanguage()` inside the
cold-activate AND warm-reactivate paths, so the fix is to give the
mobile factory a live default that reads `usePrefsStore.getState()
.voiceChatLanguage` at every call.

**#63 (CHT-016)** — `useVoiceChat.start` previously surfaced activation
failures via `Alert.alert('Voice Chat Error', ...)` — modal, no retry,
no "Open Settings" affordance. The hook now keeps the error in local
state and returns `voiceError`, `isMicPermissionError`, `retryStart()`,
and `dismissError()` so the consumer can mount its own snackbar / banner
(mirroring the existing `useUndoSnackbar` / `UploadErrorSnackbar`
pattern). `Alert.alert` is no longer called by the hook.

**#95 (STA-020)** — `setError(msg)` previously surfaced raw `lib/auth`
errors like "Sign-in callback URL is missing the required \`state\`
parameter" or "PKCE pkce_mismatch (403)". Added a local `mapSignInError`
(kept parallel with `PremiumFeatureSheet`'s helper, which can't be
imported today because the sheet is owned by a parallel PR) that maps
state / PKCE / 403, 410 / session-expired, and network / fetch shapes to
user-facing copy with a generic fallback.

## Verification
- `pnpm exec tsc --noEmit` — no new errors on touched files. (The
  pre-existing `effect`/`xstate` module-not-found errors on
  `packages/shared/src/voice-chat/*` and the worktree-only
  `usePlayerMachine` / `lib/auth` errors are unchanged by this PR.)
- Focused tests, all green:
  - `__tests__/voice-chat/service-language.test.ts` (3 tests, CHT-024)
  - `__tests__/hooks/useVoiceChat-error.test.ts` (5 tests, CHT-016)
  - `__tests__/hooks/useVoiceChat.test.ts` (7 pre-existing tests still pass)
[...]

PR #207, opened 2026-05-23
mobile(library/import): pipeline hardening cluster (#115 #122 #123 #124 #125 #129 #131)

### LINKED ISSUE #115: mobile/library/import: No duplicate-by-file-hash check on import
## Summary
Every import generates a new UUID; no hash check prevents the same file from importing twice.

## Files
- `apps/mobile/lib/file-import.ts:114-181`
- `apps/mobile/lib/sync/file-sync.ts:44-56`

## Acceptance
- Before `runImportWithService`, hash the source bytes; check books table for matching `fileHash`; return `{ ok: false, stage: 'duplicate' }` if found.
- Jest test: import same EPUB twice → second returns duplicate error.

## Domain
`mobile/library/import` — fixer batch routing.

## Source
`.parity-v3/2026-05-22-mobile/VALIDATED.md#DAT-002`

### LINKED ISSUE #122: mobile/library/import: Cover extraction failure uses opaque sentinel
## Summary
`__failed` string written to `coverPath`; no boolean column; no retry path.

## Files
- `apps/mobile/lib/book-import/adapters.ts:356`
- `apps/mobile/lib/book-storage.ts:150-168`

## Acceptance
- Add a `coverExtractionFailed: boolean` column to the schema; migrate sentinel.
- Add a "Retry cover" option in the long-press menu.

## Domain
`mobile/library/import` — fixer batch routing.

## Source
`.parity-v3/2026-05-22-mobile/VALIDATED.md#DAT-010`

### LINKED ISSUE #123: mobile/library/import: Failed import leaves orphan book directory
## Summary
Per-book dir is created before the row insert; failure to copy leaves the dir.

## Files
- `apps/mobile/lib/file-import.ts:130-140`
- `apps/mobile/lib/book-import/adapters.ts:88-117`

## Acceptance
- Wrap import in try-catch; on failure, remove the per-book directory before rethrowing.

## Domain
`mobile/library/import` — fixer batch routing.

## Source
`.parity-v3/2026-05-22-mobile/VALIDATED.md#DAT-011`

### LINKED ISSUE #124: mobile/library/import: URL import doesn't check Content-Length before
## Summary
No size check; 2GB file would OOM the device.

## Files
- `apps/mobile/lib/file-import.ts:309-371`

## Acceptance
- Read `Content-Length` header (HEAD or GET); reject if > 500MB.
- Jest test: mock response with huge size → expect throw.

## Domain
`mobile/library/import` — fixer batch routing.

## Source
`.parity-v3/2026-05-22-mobile/VALIDATED.md#DAT-012`

### LINKED ISSUE #125: mobile/library/import: Import UUID fallback can collide on rapid calls
## Summary
Fallback UUID doesn't persist state; two rapid calls could return the same ID.

## Files
- `apps/mobile/lib/file-import.ts:87-96`
- `apps/mobile/lib/file-handler.ts:42-51`

## Acceptance
- Add a counter to the fallback, OR check DB existence before insert.

## Domain
`mobile/library/import` — fixer batch routing.

## Source
`.parity-v3/2026-05-22-mobile/VALIDATED.md#DAT-013`

### LINKED ISSUE #129: mobile/library/import: URL import has no AbortController for
## Summary
Dismissing the sheet doesn't cancel the in-flight fetch; large downloads keep buffering.

## Files
- `apps/mobile/components/UrlImportSheet.tsx:26-49`
- `apps/mobile/lib/file-import.ts:309-371`

## Acceptance
- Wire an `AbortController`; pass `signal` to fetch; abort on sheet dismiss.
- Jest test: dismiss during fetch → fetch aborts.

## Domain
`mobile/library/import` — fixer batch routing.

## Source
`.parity-v3/2026-05-22-mobile/VALIDATED.md#DAT-017`

### LINKED ISSUE #131: mobile/library/import: Cover extraction is fire-and-forget; user sees
## Summary
Extraction errors are not surfaced. Users see a book with no cover with no signal that extraction failed vs is pending.

## Files
- `apps/mobile/lib/book-import/adapters.ts:403-457`
- `apps/mobile/lib/file-import.ts:151-164`

## Acceptance
- Service returns extraction status; row renders "Cover pending" or "Cover unavailable" until resolved.
- Long-press menu offers Retry.

## Domain
`mobile/library/import` — fixer batch routing.

## Source
`.parity-v3/2026-05-22-mobile/VALIDATED.md#DAT-019`

matovu-farid:
## Closes
- Closes #115 — duplicate-by-file-hash check on import (P2)
- Closes #122 — cover extraction error surfaces typed reason (P2)
- Closes #123 — failed imports clean up orphan directories (P2)
- Closes #124 — URL import checks Content-Length before download (P2)
- Closes #125 — UUID fallback collision-safe (P3)
- Closes #129 — URL import cancellable via AbortController (P3)
- Closes #131 — cover extraction surfaces completion state (P2)

## Summary
Hardening pass on the mobile book-import + URL-import pipeline. Each issue lands as a single Conventional Commits commit with TDD red→green→refactor coverage.

- **#115** — `runImportWithService` hashes the source bytes BEFORE creating the per-book dir, queries the books table for a matching `fileHash`, and short-circuits to `stage: 'duplicate'` on hit. URL import shares the same gate and tears down its tmp dir on duplicate.
- **#122** — The CoverPort now emits a typed `CoverExtractionFailureReason` discriminated union (`read-error`, `parse-error`, `write-error`, `no-cover-found`, `format-unsupported`) via a new `onExtractionFailure` callback. The persisted `'__failed'` sentinel stays for backward compat with `book-storage.mapRowToBook` (owned by sibling slot); the schema column promotion + "Retry cover" long-press menu remain follow-ups.
- **#123** — Both ownership sites for the per-book directory (`createMobileFsPort.copyBookToAppData` on the picker path and `importBookFromUrl` on the URL path) now best-effort delete the dir on failure paths (copy throw, service ok:false, service throw, duplicate hit) before re-throwing.
- **#124** — `importBookFromUrl` reads `Content-Length` from whichever response first reveals it (HEAD probe, otherwise the GET response) and rejects with a "File is too large" message above the 500 MB cap BEFORE materialising the body. Missing or non-numeric content-length falls through.
- **#125** — Both `lib/file-import.ts` and `lib/file-handler.ts` fold a module-level monotonic counter into the leading 8 nibbles of their fallback UUID, so two synchronous calls cannot collide even if `Math.random()` returns the same value back-to-back. `crypto.randomUUID` is still preferred when available.
- **#129** — `importBookFromUrl(url, { signal })` accepts an optional AbortSignal and forwards it to BOTH the HEAD probe and the GET. `UrlImportSheet` holds an `AbortController` in a ref per download and aborts it on dismiss. AbortError is suppressed in the sheet (the dismiss already tore down the UI).
- **#131** — The success outcome now carries a `coverPromise: Promise<CoverState>` that resolves to `ready` | `unavailable` | `unsupported` once the shared importer's fire-and-forget post-save pipeline finishes. The Book type, BookRow UI, and "Retry cover" affordance are owned by sibling slots; the data-layer signal is the half that unblocks them.

## Verification
- `pnpm exec tsc --noEmit` — only 4 pre-existing baseline errors in `__tests__/book-import/file-import.test.ts` (legacy `Book | null` typing, broken on `main` before this branch). 0 new errors from this PR.
- Jest: 8 affected suites pass (50+ tests new + existing):
  - `cover-extraction-failure.test.ts` (4) — sentinel unchanged
  - `cover-extraction-typed-reason.test.ts` (8) — NEW for #122
  - `cover-state-surface.test.ts` (4) — NEW for #131
  - `data-integrity.test.ts` (5) — DAT-001/004/007 unchanged
  - `duplicate-detection.test.ts` (3) — NEW for #115
  - `orphan-dir-cleanup.test.ts` (5) — NEW for #123
  - `url-import.test.ts` (15) — adjusted mocks for new DB query in #115
  - `url-import-abort.test.ts` (3) — NEW for #129
  - `url-import-size-limit.test.ts` (5) — NEW for #124
  - `uuid-fallback.test.ts` (3) — NEW for #125
  - `UrlImportSheet.test.tsx` (4) — P1-AD unchanged
  - `UrlImportSheet-abort.test.tsx` (1) — NEW for #129
- The pre-existing `book-import/file-import.test.ts` baseline failure (`Cannot use import statement outside a module` from `@react-navigation/native`) is not regressed by this PR.

## Deferred follow-ups (out of partition)
- #122: schema migration to a dedicated `coverExtractionFailed` boolean column + long-press menu for retry (touches `book-storage.ts`, `types/book.ts`, `BookRow.tsx`)
- #131: UI rendering of `coverState='pending'` / `'unavailable'` strings in BookRow (touches `BookRow.tsx`)


matovu-farid (commented):
Findings from a deep-pass review of the import-pipeline hardening cluster. (Posting as COMMENT because GitHub forbids REQUEST_CHANGES on one's own PR — verdict in summary is effectively block-ship on the first item.)

## Block ship (1)
- **Soft-deleted books permanently un-importable** — `lib/file-import.ts:226-230`. `findDuplicateByHash` queries `eq(books.fileHash, fileHash)` only — no `isDeleted` predicate. The schema has `isDeleted` (`packages/shared/src/schema.ts:31`) and the PR's own docstring claims the gate matches "non-deleted" rows. Net effect: once a user deletes a book and tries to re-import the same file, it lands as `stage:'duplicate'` with no recovery short of manual SQL.

## Important (3)
- **Picker save/parse failure leaks the per-book dir** — claim #123 says cleanup runs on "ALL failure paths". It doesn't. The shared importer (`packages/shared/src/book-import/importer.ts:175`) calls `fs.removeFile(bookPath)` on parse-fail (file only, not dir) and does nothing at all on save-fail — both leave `books/<uuid>/` behind. Only `FsPort.copyBookToAppData` throw (`lib/book-import/adapters.ts:119`) and the URL wrapper (`lib/file-import.ts:655`) actually delete the dir.
- **URL scratch dir leaks on success** — `lib/file-import.ts:622-647`. `importBookFromUrl` mints a UUID, creates `books/<urlScratchUuid>/`, writes `tmp.<format>` there, then calls `runImportWithService` which mints a DIFFERENT UUID (line 265) for the real book. On success the tmp file is deleted (line 644) but the scratch dir is never removed — every URL import leaks an empty directory whose name looks like a real book id and will confuse any future janitor walking `/books/`.
- **iOS/Android share-sheet path bypasses the dup-by-hash gate** — `lib/file-handler.ts:185`. The share-sheet flow drives the shared service directly without calling `findDuplicateByHash`. Closes-#115 should cover this surface or the issue should explicitly carve it out — right now sharing the same file twice from the OS produces two library rows.

## Polish / observability (4)
- **Sentinel string drift risk** — `COVER_EXTRACTION_FAILED_SENTINEL = '__failed'` is declared in `lib/book-import/adapters.ts:453` AND `lib/book-storage.ts:148` with comments acknowledging the duplication. Promote to a shared constant in `packages/shared` or re-export from one module — silent drift here breaks `mapRowToBook`'s sentinel detection without any test catching it.
- **Size cap relies entirely on Content-Length header** — `lib/file-import.ts:594-603`. `URL_IMPORT_MAX_BYTES = 500MB` is enforced only against the advertised header. A server that omits the header (or lies) lets a 2 GB body through `arrayBuffer()` (line 615) and OOMs anyway — exactly the bug #124 was meant to prevent. Streaming the body via a reader with a running byte total would close the loop.
- **Duplicate-detection race** — `lib/file-import.ts:256` (SELECT) and the shared service's INSERT are not in a transaction. Two concurrent imports of the same bytes both see no match and both insert. Low-probability on a phone but documented as best-effort — a unique index on `(userId, fileHash) WHERE isDeleted=0` would close it cheaply.
- **Test suite is pure-mock, not integration** — the 62 new tests stub `expo-file-system`, `@/lib/db`, `@/lib/sync/file-sync`, and the entire `createMobileBookImportService` factory. They pin contracts at the function boundary but won't catch wiring bugs (e.g. the missing `isDeleted` predicate above passes all current tests). The kanban prompt asked for integration-style tests against real fs/db — these don't meet that bar.

## Worth checking (lower confidence)
- `UrlImportSheet.tsx:74-78`: `abortRef` is cleared in `finally` only if it still points at the just-settled controller, which is correct. But `handleClose` aborts and nulls the ref synchronously; the in-flight `try` block's `controller.signal.aborted` check then suppresses both success and error branches — so the user could tap Download → cancel → tap Download again rapidly and see no spinner clearing if the first promise never reached `finally`. Race window is small but exists.
- `parseContentLengthOrNull` rejects non-integer headers, which is correct, but treats `'0'` as a valid zero-length advertisement and lets it through. The shared service will parse-fail on empty bytes, which is the desired outcome — flagging only because it's an unusual code path.

## Note
The `apps/rishi-electron/package.json` version revert (1.3.33 → 1.3.32) is a known branch-age artifact and is not counted as a finding.

## Verdict
Do not merge as-is. The `isDeleted` predicate is a one-line fix that should land before this ships; the picker-path orphan-dir gap and the share-sheet dup-bypass should either be fixed here or explicitly carved out in #115 / #123.

matovu-farid:
Thanks for the deep-pass review — patched the three blockers in three NEW commits (no force-push, no amend):

## Block ship
- **`lib/file-import.ts:226-230` — soft-deleted books permanently un-importable**
  Fixed in `e85d329c`. `findDuplicateByHash` now SELECTs with `AND isDeleted = false`. Regression covers (a) soft-deleted-only match → import proceeds, (b) live row alongside soft-deleted row → still rejected as duplicate. `apps/mobile/__tests__/book-import/duplicate-detection.test.ts`.

## Important
- **iOS/Android share-sheet bypasses dup-by-hash gate**
  Fixed in `ef34e5dc`. `findDuplicateByHash` is now exported from `lib/file-import.ts` and called from `lib/file-handler.ts` (the share-sheet entry point) before the shared service runs. Added a `'duplicate'` reason to `IncomingFileResult`. Regression in `apps/mobile/__tests__/file-handler.test.ts` asserts the gate fires for share-sheet and the import service is not invoked on a hit.
- **URL scratch dir leaks on success**
  Fixed in `daaa2b4c`. After `tmpFile.delete()` on URL-import success the scratch `books/<urlScratchUuid>/` is also removed via `safeDeleteDirectory(bookDir)`. The orphan-dir-cleanup test that previously asserted the dir stayed on success now asserts it's removed.

## Carved out
- **`packages/shared/src/book-import/importer.ts:175` — picker-path doesn't remove bookDir on parse/save fail**
  Filed as #209 with labels `critic-sweep`, `priority:p2`, `mobile:data`, assigned, and added to the Rishi project board in Todo. This is outside this PR's file partition (`packages/shared/`) so cannot be fixed here.

## Polish / observability (not in this batch — would be follow-up tickets if you want them tracked)
- Sentinel-string drift between `lib/book-import/adapters.ts:453` and `lib/book-storage.ts:148`.
- 500MB cap relies on Content-Length header (streaming reader would close the loop).
- Duplicate-detection race (no transaction wrapping SELECT + INSERT).
- Mock-only test suite (kanban prompt asked for integration-style tests).

Happy to file any of the polish items as follow-up issues if you want them tracked on the board.

Verification: `pnpm exec jest --testPathPatterns="book-import|file-handler"` → 11/12 suites pass, 78/78 tests pass. The one failing suite (`__tests__/book-import/file-import.test.ts`) fails identically on the pre-patch HEAD (`@react-navigation/native` ESM/Jest config issue) — not introduced by this patch. `pnpm exec tsc --noEmit` shows no new errors in the import-pipeline partition.

matovu-farid (commented):
3-layer code review funnel: **PASS** (review-clean; auto-merge BLOCKED by systemic CI). 0 surviving findings at Layer 1 (~10 NITs across 3 reviewers on the largest PR in this batch; 0 reached majority KEEP).

**Why clean — 7-issue import pipeline hardening cluster verified:**
- **#115** file-hash dedup: SHA-256; soft-delete predicate (`isDeleted = false`) so soft-deleted books can be re-imported. All 3 surfaces (picker, URL, share-sheet) gated.
- **#122** typed cover failures: discriminated union `CoverPortDeps.onExtractionFailure` plumbed through; `reportFailure` swallow-wraps callback so throwing observer can't break pipeline.
- **#123** orphan cleanup: picker + URL paths covered (including URL scratch-dir success cleanup).
- **#124** Content-Length pre-check: strict `/^\d+$/` parser; HEAD pre-check with signal; rejects before GET; 500MB cap.
- **#125** UUID fallback: counter increments BEFORE template fill, mixed into first 8 nibbles MSB-first; pathological `Math.random=()=>0.5` test pins distinctness; v4 variant bits preserved.
- **#129** abort: AbortController forwarded to BOTH HEAD and GET; `UrlImportSheet` uses ref (not closure) for controller; `handleClose` aborts unconditionally; post-await race guard; no error banner on user-initiated abort.
- **#131** cover state surface: `createDeferred<CoverState>` with `settled` flag prevents double-resolve.

All `catch` blocks contain `/* best-effort */` + surrounding comments. No `as any` outside test mocks. No `.skip`, `@ts-ignore`, `setTimeout(_,0)`, or `--no-verify`. PR #154 consumer-wiring precedent satisfied for the abort UX deliverable.

Non-blocking observations: `findDuplicateByHash` swallows DB errors silently (observability gap); `file-handler.ts handleIncomingFile` doesn't clean `bookDir` on `service.importFromPath` failure (partial gap vs PR's "both ownership sites" claim); `URL_IMPORT_MAX_BYTES` bypassed when both HEAD and GET omit `Content-Length` (chunked encoding — no post-`arrayBuffer()` re-check); UUID generator duplicated in two files. All deferred.

Posted as COMMENT — GitHub blocks self-approval. See .parity/codereview-207/FINDINGS.md.

PR #215, opened 2026-05-23
fix(electron/auth): cancel in-flight polls atomically with signOut

### LINKED ISSUE #161: electron/auth: in-flight polls not cancelled atomically with signOut
## Summary
Poll loops are fire-and-forget. A slow `/desktop/poll` response can land after sign-out and re-write the session. `auth-service.ts:134–191` awaits fetch outside the `cancelled` check; `writeSession` at line 172 is not gated by `!cancelled`.

## Files
- apps/rishi-electron/src/main/auth/auth-service.ts (lines 134–191)

## Acceptance
- [ ] Unit: race a poll's 200 response against `signOut()` (mocked slow fetch). `writeSession` is either NOT called or is immediately overridden by `clearSession`.
- [ ] Implementation: AbortController on the fetch, OR re-check `cancelled` after every `await`.

## Source
.parity/2026-05-23-electron/CRITIQUE-auth-deeplink.md#AUTH-008

matovu-farid:
Closes #161.

## Summary
- signOut now aborts the in-flight OAuth poll's AbortController BEFORE clearing the token store, so a late-arriving poll response can't un-sign-out the session.
- Poll callback short-circuits if its AbortSignal is already aborted (checked after every `await`).

## Test plan
- [x] `pnpm test src/main/auth/auth-service` — passes (3 new tests)
- [x] `pnpm typecheck` — clean

### DIFF
--- a/apps/rishi-electron/src/main/auth/auth-service.test.ts
+++ b/apps/rishi-electron/src/main/auth/auth-service.test.ts
@@ -1,0 +1,152 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
+
+// auth-service.ts imports `electron` (shell, BrowserWindow). Stub it so the
+// module loads under vitest. Mirrors `queries.test.ts`.
+vi.mock('electron', () => ({
+  shell: { openExternal: vi.fn().mockResolvedValue(undefined) },
+  BrowserWindow: { getAllWindows: () => [] },
+  ipcMain: { handle: () => {} },
+  app: { on: () => {}, getPath: () => '/tmp' }
+}))
+
+// We stub the session-store so we can observe writes vs clears in tests
+// without touching the filesystem or safeStorage.
+const sessionStoreMock = vi.hoisted(() => ({
+  readSession: vi.fn<() => Promise<string | null>>().mockResolvedValue(null),
+  writeSession: vi.fn<(token: string) => Promise<void>>().mockResolvedValue(undefined),
+  clearSession: vi.fn<() => Promise<void>>().mockResolvedValue(undefined)
+}))
+
+vi.mock('./session-store', () => sessionStoreMock)
+
+type FetchArgs = [input: RequestInfo | URL, init?: RequestInit]
+type FetchImpl = (...args: FetchArgs) => Promise<Response>
+
+const originalFetch = global.fetch
+
+function jsonResponse(status: number, body: unknown): Response {
+  return new Response(JSON.stringify(body), {
+    status,
+    headers: { 'Content-Type': 'application/json' }
+  })
+}
+
+describe('AuthService — signOut cancels in-flight polls atomically', () => {
+  beforeEach(() => {
+    vi.resetModules()
+    sessionStoreMock.readSession.mockReset().mockResolvedValue(null)
+    sessionStoreMock.writeSession.mockReset().mockResolvedValue(undefined)
+    sessionStoreMock.clearSession.mockReset().mockResolvedValue(undefined)
+  })
+
+  afterEach(() => {
+    global.fetch = originalFetch
+    vi.restoreAllMocks()
+  })
+
+  it('does not write a token to the session store when signOut runs while a poll is in-flight', async () => {
+    // Resolver lets the test hold the /desktop/poll response open until after signOut.
+    let resolvePoll: (res: Response) => void = () => {}
+    const pollResponsePromise = new Promise<Response>((r) => {
+      resolvePoll = r
+    })
+
+    const fetchImpl: FetchImpl = vi.fn(async (input) => {
+      const url = typeof input === 'string' ? input : input.toString()
+      if (url.endsWith('/desktop/start')) {
+        return jsonResponse(200, { state: 'state-1' })
+      }
+      if (url.endsWith('/desktop/poll')) {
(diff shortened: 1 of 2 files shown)


matovu-farid (commented):
## 3-layer code review funnel — **VERDICT: REQUEST_CHANGES** (1 blocker, 1 nit)

Reviewed against head SHA \`25eb0a3212035d647a67b84d5f743087a92f773b\`.

### Overall
The core fix is correct and well-targeted. Swapping the prior \`let cancelled\` flag for an \`AbortController\` with three post-\`await\` \`abort.signal.aborted\` gates (catch branch, between fetch and dispatch, between \`res.json()\` and \`writeSession\`) plus \`signal: abort.signal\` plumbed into the fetch closes the race correctly. Test 2 captures \`init.signal\` and asserts \`.aborted === true\` after \`signOut()\` — pins both the plumbing and the abort propagation.

**Unlike wave 1/2 PRs, this diff actually introduces NEW lint errors on its own added lines.**

### Blocker

**215-001 — Diff introduces 3 new \`@typescript-eslint/no-unnecessary-condition\` errors at the new abort checks** — \`apps/rishi-electron/src/main/auth/auth-service.ts:154, :168, :185\` (all PR-added lines)

Eslint config at \`apps/rishi-electron/eslint.config.mjs:174-177\` configures this rule as \`'error'\`. CI lint exits 1 *before* the vitest step runs, so the PR description's \"tests pass\" claim is unverifiable in CI. The lint rule is overly strict here — TS can't see that \`abort.abort()\` happens via the externally-stored \`handle.cancel\` closure under the Map's value lifetime — but it's still configured as error, so CI fails.

Verified by:
- Pulling the three error lines from \`gh run view 26333700155 --log-failed\`.
- Confirming the same line numbers exist in \`git show 25eb0a32:apps/rishi-electron/src/main/auth/auth-service.ts\`.
- Confirming \`git show main:...\` has no \`abort.signal.aborted\` references (these errors did NOT exist on main; this PR added them).

**Pick one fix:**

1. Alias the signal through a typed reference so TS no longer narrows it:
\`\`\`ts
const signal: AbortSignal = abort.signal
// then: if (signal.aborted) { ... }  in all three places
\`\`\`
2. Or three localised disables (repo already uses this style for the surrounding no-await-in-loop disables):
\`\`\`ts
// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- TS can't see external abort() via handle.cancel
if (abort.signal.aborted) { ... }
\`\`\`

Option 1 is cleaner.

### Nit

**215-002 — Test 3 (\"idempotent\") under-asserts** — \`apps/rishi-electron/src/main/auth/auth-service.test.ts:142-157\`
The test calls \`signOut()\` twice and asserts \`writeSession.not.toHaveBeenCalled()\` but does not assert \`clearSession.toHaveBeenCalledTimes(2)\`. Add the second assertion to actually pin the idempotency.

### CI note
Beyond the 3 PR-introduced errors above, the rest of the 681-error lint failure is pre-existing main debt (BookDiscoveryModal, ipc/highlights, renderer components) — same as wave 1/2 PRs. Only the 3 errors at \`auth-service.ts:154/168/185\` are attributable to this PR.

_See \`.parity/codereview-215/FINDINGS.md\` for full audit trail (6 items dropped per false-positive patterns)._

matovu-farid (commented):
## 3-layer code review funnel (re-review pass) — **VERDICT: PASS** (0 blockers, 0 nits surviving)

Re-reviewed against head SHA `46710f77261c65a4ab03aa3133a10ec9f1b5259e`.

### Funnel stats
- Layer 1 raised: 1 NIT (under-asserted idempotency test) — only 1/3 reviewers surfaced
- Layer 2 / Layer 3: vacuously pass (no majority-keep findings)

### Prior blocker verification
**215-001 (3 `no-unnecessary-condition` errors at new abort checks) — FIXED.** The fix at `apps/rishi-electron/src/main/auth/auth-service.ts:145` uses function-call indirection:

```ts
const isAborted = (): boolean => abort.signal.aborted
```

This defeats TS's control-flow analysis — TS treats the return as `boolean` at every call site (148/169/183/200), so the post-await checks are no longer narrowed to literal `false`. The `while (true)` form is explicitly permitted by `allowConstantLoopConditions: true` in `eslint.config.mjs:174-177`. A 9-line `// why:` comment at lines 136-144 documents the rationale.

This is a **real fix** — NOT a shortcut. No `eslint-disable-next-line`, no `@ts-ignore`, no `@ts-expect-error`, no `any` casts.

### Quality checks
- `pnpm lint`: 0 errors, 43 warnings (all pre-existing, none on PR-touched code)
- `pnpm typecheck`: clean
- `pnpm vitest run auth-service.test.ts`: 3/3 pass
- 8/8 no-shortcuts detection patterns: 0 hits. The lone `setTimeout(r, 0)` in the test file is a legitimate microtask-yield for AbortSignal capture, not a production race workaround.
- Ordering invariant in `signOut`: `cancelAllPolls()` runs synchronously BEFORE any await, so the post-await abort checks are load-bearing and correct.
- CI: green on head (Electron + Web + Mobile + Worker)

Posted as COMMENT (not APPROVE) because GitHub blocks self-approval. See `.parity/codereview-215/FINDINGS.md` for full vote breakdown.

PR #210, opened 2026-05-23
fix(electron/reader): debounce EPUB location-save IPC writes

matovu-farid:
Closes #174.

## Summary
- Debounce per-page-turn location-save IPC calls with a trailing flush on unmount so rapid page-turns don't hammer disk while preserving the last read position.

## Test plan
- [ ] `pnpm test -- useDebouncedLocationSave` — debounce contract test passes
- [ ] `pnpm typecheck` — clean

PR #218, opened 2026-05-23
fix(electron/db): drop cover BLOB from list queries; lazy-load via getCover

### LINKED ISSUE #190: electron/fs: lazy-load book covers (getAllBooks should not include BLOBs)
## Summary
`SELECT *` in `getAllBooks` (`queries.ts:122–126`) includes the cover BLOB for every book. 1000 books × 50 KB ≈ 50 MB per library load, all over IPC.

## Files
- apps/rishi-electron/src/main/database/queries.ts (lines 122–126)

## Acceptance
- [ ] Unit: with 1000 synthetic books × 50 KB covers, `getAllBooks()` heap delta < 5 MB. Covers fetched via a separate `cover:get(bookId)` lazy call.
- [ ] Library list IPC payload below an explicit threshold.

## Source
.parity/2026-05-23-electron/CRITIQUE-file-io.md#FIO-016

matovu-farid:
Closes #190.

## Summary
- `getAllBooks` (and the underlying `books:getAll` IPC) now SELECTs an explicit column list that **excludes the `cover` BLOB**. Cover bytes default to `[]` on returned rows. With N books in a library at ~50 KB per cover, this drops N × 50 KB of memory + IPC payload on every library refresh.
- New `getCover(bookId)` query + `books:getCover` IPC channel returns the BLOB on demand (or `null` for unknown ids). Wired through `ipc-contract.ts`, the preload bridge, `types.ts`, `test-setup.ts`, and renderer `lib/api.ts`.
- Consumer strategy: **(a) — lazy fetch**. The only renderer call site reading `book.cover` is `FileComponent.tsx`'s `BookCoverImage`. It now calls `getCover(book.id)` on mount via a coalescing in-flight map, then caches the resulting blob URL at module scope so remounts paint synchronously without flash. Total renderer delta: ~70 lines, well under the 100-line threshold for option (a).
- Existing module-level URL/Image cache behavior (warm-paint on library remount, single decode per book) is preserved; the only change is that the first paint per book now goes through an IPC fetch instead of using bytes pre-loaded by the list query.

## Tests
- Added 6 unit tests in `queries.test.ts`:
  - `_getAllBooksWithDb` returns books with `cover: []` (BLOB not loaded).
  - `_getAllBooksWithDb` still populates title / coverKind / format metadata.
  - `_getAllBooksWithDb` skips soft-deleted books (parity check).
  - `_getCoverWithDb` returns raw bytes for a known id.
  - `_getCoverWithDb` returns `null` for an unknown id.
  - `_getCoverWithDb` returns an empty Buffer when the book has no cover stored.
- Test schema was missing `file_size` (pre-existing latent gap). Added so the explicit SELECT column list resolves.

## Test plan
- [x] `pnpm exec vitest run src/main/database/queries.test.ts` — 18/18 pass (12 existing + 6 new).
- [x] `pnpm exec vitest run src/renderer/src/lib/api.test.ts src/renderer/src/components/FileComponent.test.tsx` — 19/19 pass.
- [x] `pnpm test` — 1347/1347 pass.
- [x] `pnpm typecheck` — clean.
- [x] `pnpm lint` — no new errors introduced (only pre-existing warnings remain on touched files).

## Notes
- Needed to run `pnpm rebuild better-sqlite3` once in the worktree to clear the NODE_MODULE_VERSION 140 vs 127 ABI mismatch (seen by earlier slots).
- `getBook(id)` (single-book lookup) still does `SELECT *` and returns the cover inline — intentional: reader UI needs the full row.
- `saveBook`, `updateBookCover`, sync paths are unchanged.

## Follow-ups (not in this PR)
- Could also exclude the cover from `findBookByHash` (also single-row, but hot path on import — keeping consistent for now).
- A heap-delta benchmark with 1000 synthetic books would empirically confirm the < 5 MB acceptance criterion from the issue; the structural change makes it true by construction, so deferred.

### DIFF
--- a/apps/rishi-electron/src/main/database/queries.test.ts
+++ b/apps/rishi-electron/src/main/database/queries.test.ts
@@ -12,7 +12,9 @@ import {
   _findBookByHashWithDb,
   _getBookFilepathsWithDb,
   _updateBookLastParagraphWithDb,
-  _getBookByIdWithDb
+  _getBookByIdWithDb,
+  _getAllBooksWithDb,
+  _getCoverWithDb
 } from './queries'
 
 const SCHEMA = `
@@ -21,6 +23,7 @@ const SCHEMA = `
     kind TEXT, cover BLOB, title TEXT, author TEXT, publisher TEXT,
     filepath TEXT, location TEXT, cover_kind TEXT, version INTEGER,
     sync_id TEXT, file_hash TEXT, file_r2_key TEXT, cover_r2_key TEXT,
+    file_size INTEGER DEFAULT 0,
     format TEXT, current_cfi TEXT, current_page INTEGER, user_id TEXT,
     sync_version INTEGER, is_dirty INTEGER, is_deleted INTEGER DEFAULT 0,
     last_paragraph TEXT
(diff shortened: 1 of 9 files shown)


matovu-farid (commented):
## 3-layer code review funnel (first-time review) — **VERDICT: PASS** (0 blockers, 3 nits documented inline)

Reviewed against head SHA `7d275dfe818f036f3891e387910273e4119b5493` (first review for this PR).

### Funnel stats
- Layer 1: 3 reviewers, all APPROVE
- Layer 2/3: vacuously pass for blockers; surviving nits below bias-toward-DROP threshold

### Risks cleared
- **Renderer regression:** Only `FileComponent.tsx` reads `book.cover` from list results — grep-verified across the whole renderer. It now uses the lazy `books:getCover` path.
- **N+1 queries:** Yes, one cheap row-by-id IPC per book card on first mount, coalesced via `coverFetchInFlight` Map. Intended.
- **Caching:** Module-level `coverUrlCache` + `coverImageCache` preserved exactly as pre-PR.
- **Deletion race:** `revokeCachedCoverUrl` also clears `coverFetchInFlight` — handled.
- **#210 intersection:** None. #210 modified the renderer debounce hook; #218's edits to `main/ipc/books.ts` are additive.
- **Schema parity:** Verified column-by-column — all 22 non-cover columns in `books` are in `BOOK_LIST_COLUMNS`; only `cover` omitted.
- **IPC contract:** `number[] | null` distinguishes "no book" from "no cover" coherently end-to-end.
- **CI:** all 4 jobs green

### Informational nits (not blocking)
1. **`loadCoverUrl` negative-cache comment is aspirational** — comment promises a negative cache for coverless books but code returns null without populating `coverUrlCache` (typed `Map<number,string>`, can't hold null). Result: one cheap IPC per coverless book per TanStack refetch. Comment/code mismatch only.
2. **`BookCoverImage` useEffect uses `[book]` not `[book.id]`** — slight identity-churn dependency, no IPC cost for cached books. Style nit.
3. **Delete-mid-fetch URL leak (bounded)** — if a book is deleted while its cover fetch is in flight, the in-flight promise can resurrect the cache entry after `revokeCachedCoverUrl` clears it. At most one stale entry; not user-observable.

### Quality checks
- 8/8 no-shortcuts detection patterns: 0 hits in added lines (the two pre-existing `eslint-disable` lines in FileComponent.tsx are unchanged context with proper `// why:` justifications)
- New IPC handler returns `Array.from(buf)` consistent with existing `rowToBook` marshalling convention
- Test fixture schema correctly adds `file_size INTEGER DEFAULT 0` column

Posted as COMMENT (GitHub blocks self-approval).

Issue #221, opened 2026-05-23
Worker CI failing on main — setup-bun cannot fetch tags (401 Unauthorized)


matovu-farid:
## Problem

\`Worker (typecheck)\` is failing on main HEAD (\`74284feb\`, run [26328855781](https://github.com/matovu-farid/rishi-monorepo/actions/runs/26328855781)) — but the failure is **NOT a typecheck error**. It's an infrastructure failure during job setup:

\`\`\`
Error: Failed to fetch url https://api.github.com/repos/oven-sh/bun/git/refs/tags. (status code: 401, status text: Unauthorized)
\`\`\`

The \`setup-bun\` action (used to install the Bun runtime for the Worker typecheck job) is hitting the GitHub API to resolve the latest tag and getting a 401. The job exits before \`tsc --noEmit\` even runs, so Worker code state is unverified.

## Likely causes (in order)

1. **GitHub API token in the workflow lacks read access to public repos** (e.g. \`GITHUB_TOKEN\` was scoped down recently, or \`setup-bun\` is missing a token input). Public repo tag fetches normally work unauthenticated, so a 401 specifically (not 403 rate-limit) suggests an explicitly-invalid token is being sent.
2. **Workflow permissions misconfigured.** Some GitHub Actions runners pass an empty or invalid \`GITHUB_TOKEN\` when \`permissions:\` is set too restrictively in the workflow YAML.
3. **\`setup-bun\` action regression.** The action's tag-resolution code may have started requiring explicit auth in a recent version. Check \`oven-sh/setup-bun\`'s recent releases — if pinned to \`@latest\` or a major tag, it could have picked up a breaking change.
4. **(Less likely) GitHub's tag endpoint changed behaviour** — would affect many other repos too.

## Proposed investigation

1. Open \`.github/workflows/ci.yml\` and find the Worker job's setup-bun step.
2. Check what version of \`setup-bun\` is pinned. If unpinned or \`@v1\`, pin to a known-good SHA.
3. Check the \`permissions:\` block on the workflow / job. The Worker job likely needs \`contents: read\` and \`actions: read\` for setup-bun to fetch tags.
4. If still failing, pin the Bun version explicitly via \`with: bun-version: <x.y.z>\` so the action skips tag resolution entirely.

## Acceptance

- \`Worker (typecheck)\` passes on main HEAD.
- The actual \`tsc --noEmit\` step has a chance to run and either passes or surfaces real type errors.

## Out of scope

[...]

Issue #220, opened 2026-05-23
Mobile CI failing on main — useCallback called conditionally (2 errors)


matovu-farid:
## Problem

\`Mobile (lint)\` is failing on main HEAD (\`74284feb\`, run [26328855781](https://github.com/matovu-farid/rishi-monorepo/actions/runs/26328855781)) with 2 \`react-hooks/rules-of-hooks\` errors:

\`\`\`
591:32  error  React Hook \"useCallback\" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?
681:32  error  React Hook \"useCallback\" is called conditionally. Did you accidentally call a React Hook after an early return?
\`\`\`

Both errors are at column 32, lines 591 and 681 — same file, very likely \`apps/mobile/components/HighlightsSheet.tsx\` (referenced as the culprit in prior code reviews on the electron PR queue).

## Why this matters

- Blocks any mobile PR from passing CI cleanly.
- Rules-of-hooks violations are genuine bugs — calling a hook conditionally causes hook ordering to drift between renders, which in production typically manifests as state crossing between hook slots and unpredictable re-render behaviour.

## Proposed fix

Find the early-return statements between the top of the function and lines 591/681 and either:

1. Lift the affected \`useCallback\` hooks above the early returns, OR
2. Refactor the early-return logic so all hooks run unconditionally before any return statement.

Standard React Hook rule fix — should be a small targeted patch.

## Acceptance

- \`Mobile (lint)\` passes on main HEAD.
- No \`useCallback\`/\`useEffect\`/\`useMemo\` calls after any \`return\` statement in the affected component.

## Out of scope

- The Electron CI failure (separate issue).
- The Worker CI failure (separate issue).
- The 20 mobile lint warnings (non-blocking; cleanup can follow).

Issue #219, opened 2026-05-23
Electron CI failing on main — 65 lint errors blocking PR queue


matovu-farid:
## Problem

\`Electron (lint + test + build)\` is failing on main HEAD (\`74284feb\`, run [26328855781](https://github.com/matovu-farid/rishi-monorepo/actions/runs/26328855781)) — and has been red on the last 3 commits (\`74284feb\`, \`e9d52ca1\`, \`c644ffce\`). Lint exits 1 before tests or build run.

This is **blocking the entire electron PR queue** (10 open PRs: #29, #210–#218) — they all inherit the same red CI from main, so reviewers and CI gates cannot tell signal from noise.

## Scope

- **80 files affected** across e2e tests, IPC main, renderer components, hooks, services, stores.
- **675 problems: 65 errors + 610 warnings.**
- **19 errors and 570 warnings are auto-fixable** with \`pnpm exec eslint --fix\` (or \`pnpm exec prettier --write\` for the prettier-only ones).
- **~46 errors require manual fixes** — mostly type narrowing and the new \`react-hooks/refs\` rule.

## Rule breakdown (top 12)

\`\`\`
564  prettier/prettier                                          (auto-fix)
 22  @typescript-eslint/no-unnecessary-condition                (manual)
  9  @typescript-eslint/prefer-optional-chain                   (auto-fix)
  8  react-hooks/refs                                           (manual — \"Cannot access refs during render\")
  6  @typescript-eslint/no-non-null-assertion                   (manual)
  5  @typescript-eslint/no-unnecessary-type-assertion           (auto-fix-ish)
  4  react/jsx-no-leaked-render                                 (manual)
  4  react-hooks/exhaustive-deps                                (manual)
  4  @typescript-eslint/no-empty-function                       (manual)
  3  @typescript-eslint/prefer-nullish-coalescing               (auto-fix)
  3  @typescript-eslint/no-explicit-any                         (manual)
  2  react-hooks/set-state-in-effect                            (manual)
\`\`\`

Plus 1-2 each of: \`no-import-type-side-effects\`, \`no-floating-promises\`, \`no-base-to-string\`, \`prefer-promise-reject-errors\`, \`require-await\`, \`use-unknown-in-catch-callback-variable\`, \`react-refresh/only-export-components\`, \`consistent-type-imports\`.

## Affected files (high-error density)

The 80-file list is too long for the description; full list at the CI log linked above. Highest-density files (need a careful manual pass):

[...]

PR #222, opened 2026-05-23
chore(electron): auto-fix lint pass on main (#219 part 1/2)

matovu-farid:
## What

Mechanical auto-fix pass against #219 — \`pnpm exec eslint . --fix\` followed by \`pnpm exec prettier --write .\`, no hand edits.

## Numbers

| Stage | Problems | Errors | Warnings |
|---|---|---|---|
| \`main\` HEAD (74284feb) | 675 | **65** | 610 |
| This PR | 82 | **45** | 37 |
| **Cleared** | **593** | **20** | **573** |

Typecheck (\`pnpm typecheck\` = both tsc projects) passes clean. No semantic regressions from the auto-fix.

## ⚠️ This PR does NOT green CI on its own

45 lint errors remain (rule breakdown in the commit message). Lint will still exit 1. The remaining errors are real per-site fixes:

| Rule | Count | Nature |
|---|---|---|
| \`@typescript-eslint/no-unnecessary-condition\` | 22 | TS-control-flow says condition is always truthy/falsy. Each site is either a real dead branch or a stale type narrowing. |
| \`react-hooks/refs\` (\"Cannot access refs during render\") | 8 | New rule; flags \`ref.current\` reads during render. Real refactor — move into \`useEffect\` or event handlers. |
| \`@typescript-eslint/no-empty-function\` | 4 | Empty \`connect\`/\`disconnect\` methods, likely mock/stub classes. Probably just need an \`// eslint-disable-next-line\` with rationale. |
| \`@typescript-eslint/prefer-nullish-coalescing\` | 3 | Mechanical \`||\` → \`??\` on nullable types. |
| \`react-hooks/set-state-in-effect\` | 2 | New rule; real refactor — derive state during render instead of in effect. |
| 6 singletons | 6 | \`consistent-type-imports\`, \`require-await\`, \`prefer-promise-reject-errors\`, \`no-base-to-string\`, \`use-unknown-in-catch-callback-variable\`, \`no-floating-promises\`, \`react-refresh/only-export-components\` |

Follow-up issue tracks the manual cleanup.

## Why split

- **Mechanical pass is low-risk and high-volume.** Reviewing 66 files of \`eslint --fix\` + \`prettier --write\` changes is much easier when the diff is pure mechanical (no semantic intent to vet).
- **Manual fixes are high-risk and low-volume.** Each remaining error needs a per-site judgement call (is the dead branch real intent? does the ref access need to move into an effect or into a handler?). Reviewing them benefits from a small focused diff.
- **Rebaseable inheritance.** The 10 open electron PRs (#29, #210–#218) can rebase on this branch / on main once merged. Cuts down the per-PR churn from prettier (570 warnings).

## Scope of this PR

- Only \`.ts\`/\`.tsx\`/\`.css\` source files modified. 12 markdown plan docs in \`apps/rishi-electron/docs/superpowers/\` were touched by prettier — reverted because they're irrelevant to CI lint and would add noise.

## What's NOT in this PR

- The remaining 45 manual errors (tracked in the follow-up issue).
- The Mobile lint failure (#220).
- The Worker setup-bun failure (#221).
- Any changes to the lint config or rule severity — the rules are configured \`error\` intentionally as footgun-catchers.

Closes part 1 of #219.

### DIFF
--- a/apps/rishi-electron/e2e/navigation-history-epub.spec.ts
+++ b/apps/rishi-electron/e2e/navigation-history-epub.spec.ts
@@ -105,9 +105,9 @@ test.describe('Navigation history — EPUB', () => {
     const tocVisible = await tocPanel.isVisible({ timeout: 5000 }).catch(() => false)
     if (!tocVisible) {
       // Try the sheet-based ReaderTOC (used in EPUB via ReactReader).
-      await expect(
-        bookPage.locator('text=Table of Contents').first()
-      ).toBeVisible({ timeout: 5000 })
+      await expect(bookPage.locator('text=Table of Contents').first()).toBeVisible({
+        timeout: 5000
+      })
     }
 
     // Click the first TOC entry button.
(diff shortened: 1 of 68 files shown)


matovu-farid (commented):
## Code review funnel — **VERDICT: PASS** (1 LOW advisory, 0 blockers)

Reviewed against head SHA \`7d7d77575615ff3069d661fd8239a47d6e1a45d0\` (both commits).

### Overall
Exactly what the PR claims: 66 files of mechanical \`eslint --fix\` + \`prettier --write\` (commit \`ec98f35e\`) and 7 per-site singleton fixes (commit \`7d7d7757\`).

I verified each of the 7 singletons against the file at HEAD and spot-checked the auto-fixes from commit 1 (non-null assertion removals, \`&&\`→\`?.\` rewrites, redundant casts). All safe and type-equivalent. The FileReader change in \`encode.ts\` was checked against MDN — \`reader.result\` after \`readAsDataURL()\` is always a string; \`reader.error\` is always \`DOMException\` (inherits from \`Error\` per WebIDL since 2021 in Chromium/Electron). The new defensive branches are unreachable but harmless. Observable behavior identical.

### Advisory — 222-001 (LOW, advisory, NOT a blocker)

**\`EpubView.tsx:94\` \`||\` → \`??\` changes BOOK_OPENED firing for freshly-imported, never-opened books.**

Type analysis: \`book.location\` is non-nullable per the \`Book\` interface, but \`main/ipc/formats.ts:129,152\` initializes it as \`''\` for freshly-imported books. The commit message's claim that \`book.location\` is \`string|null|undefined\` is wrong on the type but right in spirit (the field IS initialized empty).

- Old (\`||\`): \`null || '' || '0'\` → \`'0'\`
- New (\`??\`): \`null ?? '' ?? '0'\` → \`''\`

Downstream: \`useEffect\` at line 957 (deps \`[book.id]\`) early-returns when \`!currentLocation\`, so \`BOOK_OPENED\` is now skipped on first mount when the book has empty \`location\` AND no \`lastParagraph\`. The \`navigationHistoryMachine\` stays in \`inactive\` for the whole reader session; all \`PAGE_VISITED\`/\`JUMP_REQUESTED\` events become no-ops.

**Net effect:** for the very first open of a freshly-imported EPUB, the navigation-history pill and stack are silently disabled for that reader session. Reopening the book after epubjs writes a CFI back to \`book.location\` (line 1049) clears the issue.

**Why advisory, not blocker:**
- The PR is explicitly mechanical and expected to land with remaining lint errors (45 → 38; tracked in #223).
- The bug surface is narrow (first-open of freshly-imported book) and self-heals on next open.
- The *old* behavior was *also* arguably wrong — it sent the literal junk CFI \`'0'\` to \`BOOK_OPENED\`, which silently corrupted the machine's \`initialPosition\`.

**Suggested follow-up issue:** change the \`BOOK_OPENED\` effect at \`EpubView.tsx:957-967\` to either depend on \`[book.id, currentLocation]\` so it re-fires when a real CFI arrives, or use a \`book.location || '0'\` fallback in the initial \`useState\` to preserve the previous nav-history bootstrapping semantics.

### CI note
PR explicitly says lint CI will still fail (45 → 38 errors). Remaining tracked in #223. Out of scope for this review.

_See \`.parity/codereview-222/FINDINGS.md\` for the full audit trail (8 spot-checked items with no findings)._

PR #224, opened 2026-05-23
fix(mobile/reader): hoist useCallback above early-returns (#220)

### LINKED ISSUE #220: Mobile CI failing on main — useCallback called conditionally (2 errors)

matovu-farid:
## Problem (#220)

\`Mobile (lint)\` was failing on main with 2 \`react-hooks/rules-of-hooks\` errors:

- \`apps/mobile/app/reader/djvu/[id].tsx:591:32\` — \`useCallback\` after the loading + error early-returns
- \`apps/mobile/app/reader/mobi/[id].tsx:681:32\` — same pattern

Rules-of-hooks violations are real bugs: when the early-return fires, the hook isn't called, but on a re-render where the conditions don't fire, the hook IS called — so React's hook-order invariant breaks, which in production can cause state to cross between hook slots.

## Fix

Hoisted both \`getActivationContext\` \`useCallback\` blocks (including their docblocks) to immediately after the \`tocItems\` \`useMemo\` — the deepest dependency. All other deps (\`pageCount\`, \`currentPage\`, \`activeParagraph\` on djvu; \`chapterCount\`, \`currentChapter\`, \`activeParagraph\` on mobi) are also declared earlier in the function, so the hoist is safe.

Updated the docblock to note: \"Declared above the loading/error early-returns so the hook order stays stable across renders (react-hooks/rules-of-hooks).\" Anyone touching this code later sees the constraint.

No other behavioural change. Closure shape, dependency arrays, and consumer call site (\`<ReaderShell getActivationContext={getActivationContext} />\`) are unchanged.

## Verification

\`\`\`
\$ cd apps/mobile && pnpm lint
✖ 20 problems (0 errors, 20 warnings)
\`\`\`

vs main HEAD:

\`\`\`
\$ cd apps/mobile && pnpm lint
✖ 22 problems (2 errors, 20 warnings)
\`\`\`

Net change: -2 errors, warnings unchanged. The 20 remaining warnings are pre-existing main-branch lint debt (Array<T> vs T[], no-unused-expressions, etc.) and don't affect the CI gate.

\`pnpm exec tsc --noEmit\` from the mobile dir surfaces pre-existing dep-resolution errors in \`packages/shared/src/voice-chat/*.ts\` (\`xstate\`, \`effect\` not resolvable from mobile's tsconfig); these are unrelated to this PR — the mobile CI job only runs \`lint\`, not \`typecheck\`.

Closes #220.

### DIFF
--- a/apps/mobile/app/reader/djvu/[id].tsx
+++ b/apps/mobile/app/reader/djvu/[id].tsx
@@ -489,6 +489,25 @@ export default function DjvuReaderScreen() {
     }))
   }, [pageCount])
 
+  // CHT-002 — voice-chat activation context. DJVU is canvas-only (no text
+  // layer that survives the WebView round-trip), so the best we can do is
+  // pass a "Page N of M" page-text proxy plus the synthesised per-page
+  // outline. When TTS is active, forward the active paragraph too — the
+  // shared extractor can light this up if the DJVU has an OCR layer.
+  // Declared above the loading/error early-returns so the hook order stays
+  // stable across renders (react-hooks/rules-of-hooks).
+  const getActivationContext = useCallback(() => {
+    const outline = tocItems.map((t) => ({ href: t.href, label: t.label }))
+    const pageText = pageCount > 0
+      ? `Page ${currentPage} of ${pageCount}`
+      : `Page ${currentPage}`
+    return {
+      pageText,
+      outline,
+      activeParagraphText: activeParagraph?.text ?? undefined,
+    }
+  }, [tocItems, currentPage, pageCount, activeParagraph])
+
   const currentTocHref = useMemo(
     () => `djvu-page:${currentPage}`,
     [currentPage],
@@ -582,23 +601,6 @@ export default function DjvuReaderScreen() {
     pageCount > 0
       ? { kind: 'page', current: currentPage, total: pageCount }
       : { kind: 'none' }
-
-  // CHT-002 — voice-chat activation context. DJVU is canvas-only (no text
-  // layer that survives the WebView round-trip), so the best we can do is
-  // pass a "Page N of M" page-text proxy plus the synthesised per-page
-  // outline. When TTS is active, forward the active paragraph too — the
-  // shared extractor can light this up if the DJVU has an OCR layer.
-  const getActivationContext = useCallback(() => {
-    const outline = tocItems.map((t) => ({ href: t.href, label: t.label }))
-    const pageText = pageCount > 0
-      ? `Page ${currentPage} of ${pageCount}`
-      : `Page ${currentPage}`
-    return {
-      pageText,
-      outline,
-      activeParagraphText: activeParagraph?.text ?? undefined,
-    }
-  }, [tocItems, currentPage, pageCount, activeParagraph])
 
   const djvuNavCluster = (
     <DjvuNavCluster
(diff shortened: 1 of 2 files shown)


matovu-farid (commented):
## Code review funnel — **VERDICT: PASS** (0 findings)

Reviewed against head SHA \`5db88a7f926804296a52add43f5fcc72bc1aa761\`.

### Overall
Textbook minimal lint fix. The PR does exactly what the description says: hoists \`getActivationContext\` \`useCallback\` above the loading/error early-returns in both \`djvu/[id].tsx\` and \`mobi/[id].tsx\` to satisfy \`react-hooks/rules-of-hooks\`.

### Verification

1. **Hoist actually fixes rules-of-hooks?** ✅ DJVU \`useCallback\` now at line 499, both early-returns at lines 577 and 588. MOBI \`useCallback\` at line 485, early-returns at lines 664 and 675. No \`return\` between function-start and the new hook locations.
2. **All deps declared earlier?** ✅
   - DJVU deps: \`pageCount\` (line 212), \`currentPage\` (213), \`activeParagraph\` (238), \`tocItems\` (484) — all before line 499.
   - MOBI deps: \`currentChapter\` (275), \`activeParagraph\` (302), \`tocItems\` (469) — all before line 485.
3. **Inadvertent reordering?** ✅ None. Diff is pure positional move; closure body, dep array, and consumer call site (\`<ReaderShell getActivationContext={getActivationContext} />\`) are byte-identical between old and new locations.
4. **Docblock preserved?** ✅ Plus one added line documenting the constraint (\"Declared above the loading/error early-returns so the hook order stays stable across renders (react-hooks/rules-of-hooks)\") — net-positive for the next reader.

Diff stat (+36/-32 across 2 files) matches \`PR.json\`. One commit, matches head SHA.

### CI note
PR brings \`Mobile (lint)\` from 2 errors to 0 (warnings unchanged at 20). This closes the mobile CI failure that was blocking issue #220.

_See \`.parity/codereview-224/FINDINGS.md\` for the full audit trail._

Issue #177, opened 2026-05-23
electron/reader: surface iframe-detached error from resolve-live-selection


matovu-farid:
## Summary
`views.forEach()` iterates visible views; `contents.window?.getSelection()` fails silently if the iframe has unloaded. Caller gets `null` and assumes no selection exists.

## Files
- apps/rishi-electron/src/renderer/src/modules/resolve-live-selection/index.ts (lines 44–62)

## Acceptance
- [ ] Unit: detached iframe → resolver returns a sentinel (e.g. `{ error: 'detached' }`) and warning is logged.
- [ ] Callers can distinguish 'no selection' from 'selection unavailable'.

## Source
.parity/2026-05-23-electron/CRITIQUE-reader.md#RDR-008

matovu-farid:
Closing: same reason as #173 — this is about epub.js's view lifecycle (iframe unloads between our query and the handler). The 'fix' would only surface the error from our wrapper, not prevent it. Not worth a band-aid PR; revisit if epub.js exposes a stable iframe lifecycle hook.

PR #225, opened 2026-05-23
fix(ci): pin setup-bun to SHA + explicit bun-version (#221)

### LINKED ISSUE #221: Worker CI failing on main — setup-bun cannot fetch tags (401 Unauthorized)

matovu-farid:
## Problem (#221)

\`oven-sh/setup-bun@v2\` was failing on every CI run with:

\`\`\`
Error: Failed to fetch url https://api.github.com/repos/oven-sh/bun/git/refs/tags.
(status code: 401, status text: Unauthorized)
\`\`\`

When \`bun-version\` is not supplied, setup-bun's tag-resolution logic hits the GitHub API to pick \"latest\". That endpoint started returning 401, blocking the Worker typecheck job (and likely also the other 3 setup-bun usages we have).

## Fix

Sidestep the resolve entirely:

1. **Pin the action by SHA** (\`@0c5077e51419868618aeaa5fe8019c62421857d6\`). The SHA was already sitting in a trailing comment — likely the intent of a prior commit that never applied it.
2. **Specify \`bun-version: 1.1.42\`** explicitly. With a concrete version supplied, the action skips the tags-API call entirely.

Applied to all 4 setup-bun usages:

| File | Line | Job |
|---|---|---|
| \`.github/workflows/ci.yml\` | 59 | (job above the worker block — appears to be mobile/web) |
| \`.github/workflows/ci.yml\` | 93 | (next job) |
| \`.github/workflows/ci.yml\` | 121 | Worker (typecheck) |
| \`.github/workflows/deploy-worker.yml\` | 28 | Worker deploy |

## Verification

This is a CI-only change — can't verify locally. The PR's own CI run is the proof. Expected: \`Worker (typecheck)\` proceeds past setup and actually runs \`bunx tsc --noEmit\` for the first time in days, which either passes or surfaces real type errors that have been hidden behind the setup failure.

## Notes

- **Version choice (1.1.42).** The repo doesn't pin Bun anywhere (no \`.bun-version\`, no \`engines.bun\`), so this is the first explicit version pin. If something the worker uses requires a newer Bun, bump the version in a follow-up. Safe to start conservative.
- **Why both SHA + version pin.** Belt-and-braces. Pinning the action alone wouldn't help if the API call is in the action's logic (which it is). Pinning the bun-version alone would work, but leaving \`@v2\` is bad CI hygiene (the tag can be moved).
- **Other setup-bun usages updated.** \`ci.yml\` has three setup-bun lines (one per job that needs bun). Fixed all three with the same treatment, so we don't get the failure on other jobs once one is debugged.

Closes #221.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: oven-sh/setup-bun@v2  # 0c5077e51419868618aeaa5fe8019c62421857d6
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6  # pinned-by-SHA + explicit bun-version skips the GitHub tags-API resolve that returns 401 (#221)
+        with:
+          bun-version: 1.1.42
 
       - uses: actions/cache@v4
         with:
@@ -90,7 +92,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: oven-sh/setup-bun@v2  # 0c5077e51419868618aeaa5fe8019c62421857d6
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6  # pinned-by-SHA + explicit bun-version skips the GitHub tags-API resolve that returns 401 (#221)
+        with:
+          bun-version: 1.1.42
 
       - uses: actions/cache@v4
         with:
@@ -118,7 +122,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: oven-sh/setup-bun@v2  # 0c5077e51419868618aeaa5fe8019c62421857d6
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6  # pinned-by-SHA + explicit bun-version skips the GitHub tags-API resolve that returns 401 (#221)
+        with:
+          bun-version: 1.1.42
 
       - uses: actions/cache@v4
         with:
--- a/.github/workflows/deploy-worker.yml
+++ b/.github/workflows/deploy-worker.yml
@@ -25,7 +25,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: oven-sh/setup-bun@v2  # 0c5077e51419868618aeaa5fe8019c62421857d6
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6  # pinned-by-SHA + explicit bun-version skips the GitHub tags-API resolve that returns 401 (#221)
+        with:
+          bun-version: 1.1.42
 
       - uses: actions/cache@v4
         with:
--- a/.github/workflows/release-mobile.yml
+++ b/.github/workflows/release-mobile.yml
@@ -40,7 +40,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: oven-sh/setup-bun@v2  # 0c5077e51419868618aeaa5fe8019c62421857d6
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6  # pinned-by-SHA + explicit bun-version skips the GitHub tags-API resolve that returns 401 (#221)
+        with:
+          bun-version: 1.1.42
 
       - uses: actions/cache@v4
         with:


matovu-farid (commented):
## Code review funnel — **VERDICT: PASS** (1 finding addressed in follow-up commit)

Reviewed against head SHA \`91201d6ac571602391d1c3e3f69e3578136efb7f\` (initial). The review surfaced one real omission — fixed by the follow-up commit \`f09d1b5a\`, now part of this PR.

### Overall
The fix is correct and verifies against upstream:
- **SHA \`0c5077e5…\` is real** — \"release: v2.2.0 (#177)\" on \`oven-sh/setup-bun\` (verified via GitHub).
- **Bun \`1.1.42\` is real** — released 2024-12-21.
- **\`bun-version\` is the correct input key** per setup-bun's \`action.yml\`.
- Indentation is correct (\`with:\` 2 spaces deeper than \`uses:\`).
- Cache step / \`hashFiles('**/bun.lockb')\` untouched.

### Finding caught + addressed

**225-001 — Missing 5th setup-bun usage in \`release-mobile.yml:43\`** (was LOW advisory; now fixed in commit \`f09d1b5a\`)

The initial PR claimed 4 setup-bun usages; the repo actually had 5. The 5th in \`release-mobile.yml\` would have hit the same 401 on the next mobile release cut. Same fix applied — pin to SHA + explicit \`bun-version\`.

All 5 setup-bun usages now consistent:
- \`ci.yml\` x3 (lines 59, 95, 125)
- \`deploy-worker.yml\` x1 (line 28)
- \`release-mobile.yml\` x1 (line 43, **new in follow-up commit**)

### Other notes (non-findings)
- The pinned SHA is v2.2.0 of setup-bun. \`@v2\` currently floats to v2.1.3 — so this is also a minor version bump (Node 24 action runtime). Intentional and desirable hygiene.
- \`bun-version: 1.1.42\` is a sensible conservative pin. If the worker needs newer Bun features, bump in a follow-up.

### CI note
This is a CI-only change. Confirmation comes from PR's own CI run: expected \`Worker (typecheck)\` to proceed past setup and actually run \`bunx tsc --noEmit\` for the first time in days.

_See \`.parity/codereview-225/FINDINGS.md\` for the full audit trail._

Issue #226, opened 2026-05-23
EpubView: nav-history dormant for freshly-imported, never-opened books (#222 follow-up)


matovu-farid:
## Background

Follow-up from the code review on #222 (\`chore(electron): auto-fix lint pass on main\`). The PR's commit \`7d7d7757\` changed \`EpubView.tsx:94\` from \`||\` to \`??\`:

\`\`\`ts
// before
const [currentLocation, setCurrentLocation] = useState<string>(resumeCfi || book.location || '0')
// after
const [currentLocation, setCurrentLocation] = useState<string>(resumeCfi ?? book.location ?? '0')
\`\`\`

This is a mechanical \`prefer-nullish-coalescing\` lint fix. But \`book.location\` is initialized as \`''\` (empty string) for freshly-imported EPUB and PDF books by \`main/ipc/formats.ts:129,152\`. Empty string is falsy but not nullish, so the two operators behave differently:

| Inputs | \`||\` (old) | \`??\` (new) |
|---|---|---|
| \`null, '', '0'\` | \`'0'\` | \`''\` |

## Downstream impact

\`EpubView.tsx:957-967\` runs a \`useEffect\` (deps \`[book.id]\`) that early-returns when \`!currentLocation\`. Before #222, freshly-imported books got \`currentLocation = '0'\` which is truthy, so \`BOOK_OPENED\` fired into \`navigationHistoryActor\`. After #222, freshly-imported books get \`currentLocation = ''\` (falsy), so \`BOOK_OPENED\` is skipped on first mount.

\`navigationHistoryMachine.ts:104-108\` only transitions from \`inactive\` → \`active\` when it receives \`BOOK_OPENED\`. So:

- For the **very first open** of a freshly-imported EPUB, the navigation-history pill and stack are silently disabled for that whole reader session.
- Subsequent opens are fine — epubjs writes a real CFI back to \`book.location\` (\`EpubView.tsx:1049\`) on first open, so the next open has a non-empty location.
- Workaround for the user: close and reopen the book.

## But the old behaviour wasn't quite right either

The old code was sending the literal junk CFI \`'0'\` to \`BOOK_OPENED\`. \`'0'\` is not a valid epubcfi (epubjs falls back to \`String(0)\` → page-zero), so the machine's \`initialPosition\` was being silently corrupted. So **neither old nor new is fully correct.**

## Proposed fix

Two options, either works:

[...]

PR #227, opened 2026-05-23
chore(electron): manual cleanup of 38 remaining lint errors (#223)

### LINKED ISSUE #223: Electron CI: manual cleanup of 45 remaining lint errors (#219 part 2/2)
## Context

Part 2 of #219. After the auto-fix pass landed in #222 (66 files, 593 problems cleared), 45 lint errors remain that need per-site human judgement. **Lint still exits 1 on main until these are fixed** — same blocking effect on the electron PR queue.

### Type narrowing (22 — \`@typescript-eslint/no-unnecessary-condition\`)

TS control-flow says the condition is always truthy/falsy. Each site is one of:
- **Real dead branch.** The branch is unreachable; delete it. Look for sentinel checks against types that no longer include \`null\`/\`undefined\`.
- **Stale type narrowing.** A field's type was tightened (e.g. \`T\` instead of \`T | undefined\`) but the guard remained. Either widen the field's type to match runtime, or delete the guard if the field is truly always-present.

Specific patterns seen in the log:
- \`Unnecessary conditional, expected left-hand side of \`??\` operator to be possibly null or undefined\` — RHS is never reached.
- \`Unnecessary optional chain on a non-nullish value\` — \`.?.\` is redundant.
- \`Unnecessary conditional, value is always falsy\` — branch never fires.
- \`Unnecessary conditional, comparison is always true, since "X" === "X" is true\` — copy-paste bug.

### react-hooks/refs (8 — \"Cannot access refs during render\")

New rule from \`eslint-plugin-react-hooks\` v6+. Flags \`ref.current\` access in the render body (where React's concurrent renderer can't guarantee freshness). Fix per-site:
- If the ref read is used to derive JSX, capture it in a \`useLayoutEffect\` and set a state. Or compute from state directly.
- If the ref read is in a callback that's only called by an event/effect, move it into the callback closure.
- If the ref is genuinely render-time required (rare), the rule has an escape hatch via the official \`react-compiler-runtime\` annotation, but that should be a last resort.

Touch points: at least one each in EpubView.tsx (5), TTSControls.tsx (1), and other reader/highlight components (2).

### react-hooks/set-state-in-effect (2 — \"Calling setState synchronously within an effect can trigger cascading renders\")

Another new rule. The pattern \`useEffect(() => setX(...), [...])\` causes an extra render pass that React's renderer flags as wasteful. Fix per-site:
- If the state is **derived** from inputs, compute it inline during render instead of via an effect.
- If the state is for **subscription** (\`subscribe → setState → unsubscribe\`), migrate to \`useSyncExternalStore\`.
- If the effect genuinely models a side-effect that updates state (e.g. async fetch → setState), the rule has a workaround pattern documented in the React 19 docs.

Touch points: EpubView.tsx:856, somewhere else around :187 (different file).

### no-empty-function (4)

Empty \`connect\`/\`disconnect\` methods, likely on a Mock or Stub class. Two options:
1. Add a noop body comment and disable the rule on those lines:
   \`\`\`ts
   // eslint-disable-next-line @typescript-eslint/no-empty-function -- intentional noop in mock impl
   connect(): void {}
   \`\`\`
2. Add a no-op annotation: \`connect(): void { /* noop */ }\` (some configs accept this).

### Singletons (10 errors across ~7 rules)

| Rule | Fix |
|---|---|
| \`@typescript-eslint/consistent-type-imports\` | Change \`import()\` annotation to top-level \`import type\` |
| \`@typescript-eslint/require-await\` | \`applyVisual\` method declared \`async\` but has no \`await\`. Drop the \`async\` keyword or add an \`await\`. |
| \`@typescript-eslint/prefer-promise-reject-errors\` | Wrap the reject argument in \`new Error(...)\` |
| \`@typescript-eslint/no-base-to-string\` | \`reader.result\` interpolated into a template literal would stringify as \`[object Object]\`. Add explicit \`String(...)\` or type-narrow first. |
| \`@typescript-eslint/use-unknown-in-catch-callback-variable\` | Type the catch callback's variable as \`unknown\` |
| \`@typescript-eslint/no-floating-promises\` | Add \`void\` prefix, \`.catch(...)\`, or \`await\` |
| \`react-refresh/only-export-components\` | Move the non-component export to a separate file (or disable on that file with a HMR-impact comment) |
| \`@typescript-eslint/prefer-nullish-coalescing\` (3 sites) | Change \`||\` to \`??\` for nullable LHSes |

## Acceptance

- \`Electron (lint + test + build)\` passes on main HEAD.
- No suppressions at the lint config level — only per-line eslint-disable comments with a clear \`-- <rationale>\` are acceptable (and only for the truly defensible cases like mock noops).
- Typecheck (\`pnpm typecheck\`) remains clean.
- Tests still pass (\`pnpm test\`).

## Estimated effort

Smaller errors (singletons + 3 nullish-coalescing + 4 empty-function disables): ~15 minutes of mechanical work.

\`no-unnecessary-condition\` (22): ~1-2 hours, because each site needs to be read in context to decide \"delete branch\" vs \"widen type\".

\`react-hooks/refs\` (8) + \`set-state-in-effect\` (2): ~2-4 hours, because these are real refactors that need careful testing per change.

Total: half a day of focused work, or a day if conservative.

## Why this is still P0

This issue blocks the 10 open electron PRs (#29, #210–#218) from getting a clean CI signal. Reviewers can't tell PR-introduced regressions from pre-existing main debt while it's red. Resolving brings the funnel back to signal-only.

matovu-farid:
## What

Closes #223 — the manual cleanup of the 38 lint errors that survived the auto-fix pass in #222. With this PR, \`Electron (lint + test + build)\` goes from **38 errors** to **0 errors** on a freshly-rebased main. Combined with #222, the full electron lint cleanup took main from 65 → 0.

**Hard rule observed throughout: no \`eslint-disable\` shortcuts.** Every fix restructures the actual code.

## Numbers

| Stage | Errors | Warnings |
|---|---|---|
| \`main\` HEAD after #222 merge | 38 | 37 |
| Batch A (mechanical singletons) | 32 | 37 |
| Batch B (no-unnecessary-condition) | 9 | 40 |
| Batch C+D (react-hooks refactors) | **0** | 40 |

Typecheck (both tsc projects) clean. 1329 / 1329 tests pass across 146 test files. (\`src/main/database/queries.test.ts\` has 12 better-sqlite3 ABI mismatch failures that are confirmed pre-existing on main HEAD — same failures with or without these changes.)

## Real bugs caught along the way

1. **\`lib/visualHeuristic.ts:184\` dead branch** — \`else if (hit.kind === 'image')\` was unreachable. \`classifyElement\` only returns \`'equation' | 'figure'\`; standalone \`\` elements are classified as \`'figure'\` and routed to \`images++\` via an inner \`tag === 'img'\` check. Collapsed to \`else\` with a comment.
2. **\`modules/file-sync.ts:122\` unsound type narrowing** — \`data.code === 'STORAGE_LIMIT_REACHED'\` flagged as tautology. Root cause: \`data\` came from \`res.clone().json()\` and was typed as \`Partial<UploadUrlError>\`. But \`data\` is **untrusted** — a malformed worker response could carry any string. The prior typed narrowing was unsound. Restructured to read \`data.code\` through \`unknown\` so the runtime validation is honest.

## Type-widening fixes (real bugs hidden by stale narrowing)

- \`hooks/reader/useBookSyncId.ts:37,41\` — \`state.cancelled\` narrowed to literal \`false\` after first early-return. The cleanup mutation was invisible to TS. Switched to \`let cancelled\` + getter so reads cross a function boundary.
- \`stores/pdfStore.ts:220\` — \`Record<number, FooterMask>\` widened to \`Partial<Record<…>>\` so index access returns \`FooterMask | undefined\`.
- \`hooks/usePdfTextSelection.ts:55\` — \`rects[0]\` → \`rects.at(0)\` for proper \`T | undefined\`.

## Real React 19 refactors

- **EpubView.tsx** — extracted iframe tracking into \`useVisibleEpubIframe(rendition)\` using \`useSyncExternalStore\`. Subscribe handles \`rendition.on('rendered')\`; snapshot derives via \`getVisibleIframe\`. Returns null when rendition is null. Eliminates the sync setState that triggered the rule. Gains concurrent-mode tearing safety.
- **TTSControls.tsx** — replaced \`useEffect\`-based sticky state with the React docs "track input that drives derived value" pattern. Stores \`lastNonLoadingState\`; derives \`showRepeat\` during render. Conditional setState during render is the bail-out path (not cascading).
- **usePdfTextSelection.ts + pdf.tsx** — dropped \`paramsRef\` shim, destructured callbacks at hook top, added 4 \`useCallback\`s at the call site so the listener doesn't rebind on every parent render. The right fix (stable boundary callbacks), not a ref hack.
- **pdf-page.tsx** — converted \`useRef<HTMLDivElement>\` to \`useState + callback ref\`. Render reads state instead of \`.current\`; callbacks include the state in their deps. Initial render still gates HighlightLayer correctly.

## Mechanical singletons (still real fixes)

- \`pdf.tsx:755\` \`async\` → sync (\`require-await\`). The \`HighlightTarget.applyVisual\` interface already accepts \`() => Promise<void> | void\`; sync return type is valid.
- \`local-vad.test.ts\` empty mock methods → \`vi.fn()\` arrow properties. Satisfies interface AND makes them spyable.
- \`pdf-page.tsx\` \`makeCustomTextRenderer\` extracted to its own file (\`react-refresh/only-export-components\`). The function was already documented as "extracted from the component closure for unit testing"; moving it to its own module follows through on that intent.

## New file

- \`apps/rishi-electron/src/renderer/src/hooks/reader/useVisibleEpubIframe.ts\` — declarative wrapper around \`rendition.on('rendered')\` + \`getVisibleIframe\` using \`useSyncExternalStore\`. ~49 lines.

## What's NOT in this PR

- The 40 remaining lint warnings (out of scope; tracked separately if anyone cares to clean them up).
- The Mobile + Worker fixes (already merged via #224 and #225).
- The advisory finding from #222 about EpubView.tsx:94 nav-history bootstrapping (#226 — separate scope).
- Any change to the lint config or rule severity.

Closes #223. Once merged, \`Electron (lint + test + build)\` will be fully green on main, unblocking the 10 in-flight electron PRs (#29, #210–#218).

### DIFF
--- a/apps/rishi-electron/src/main/ipc/highlights.ts
+++ b/apps/rishi-electron/src/main/ipc/highlights.ts
@@ -14,12 +14,13 @@ export function registerHighlightHandlers(): void {
       .all()
 
     // Map on-disk rows to the renderer-facing HighlightRow shape:
-    // - format defaults to 'epub' for legacy rows that predate v2 migration
+    // - format is NOT NULL with a default of 'epub' at the column level, so
+    //   legacy rows already read back as 'epub' from drizzle — no fallback needed.
     // - cfi_range '' (written by PDF rows to satisfy NOT NULL) becomes null
     // - locator passes through as-is (null for EPUB rows)
     return rows.map((row) => ({
       ...row,
-      format: (row.format ?? 'epub') as 'epub' | 'pdf',
+      format: row.format as 'epub' | 'pdf',
       cfiRange: row.cfiRange === '' ? null : row.cfiRange,
       locator: row.locator ?? null
     }))
--- a/apps/rishi-electron/src/renderer/src/components/epub/EpubView.tsx
+++ b/apps/rishi-electron/src/renderer/src/components/epub/EpubView.tsx
@@ -91,7 +91,7 @@ export default function EpubView({ book }: { book: Book }): React.JSX.Element {
   // a stale id from a different format would otherwise be handed to epubjs,
   // which silently ignores it and opens at the start of the book.
   const resumeCfi = book.lastParagraph?.startsWith('epubcfi(') ? book.lastParagraph : null
-  const [currentLocation, setCurrentLocation] = useState<string>(resumeCfi ?? book.location ?? '0')
+  const [currentLocation, setCurrentLocation] = useState<string>(resumeCfi ?? book.location)
   // Sync with book.location when it changes from a refetch (e.g., returning from library).
   // Only sync before the rendition has settled to avoid overriding user navigation.
   const bookLocationRef = useRef(book.location)
@@ -451,7 +451,11 @@ export default function EpubView({ book }: { book: Book }): React.JSX.Element {
     const sync = (): void => {
       // epubjs Views exposes `first()` at runtime even though the typings
       // declare it only as `View[]`. Cast through unknown to access it.
-      const views = rendition.manager?.views as unknown as
+      // `manager` is not on the public Rendition type — go through unknown so
+      // we can probe defensively at runtime without TS treating the chain as
+      // always-defined.
+      const manager = (rendition as unknown as { manager?: { views?: unknown } }).manager
+      const views = manager?.views as
         | { first?: () => { iframe?: HTMLIFrameElement } | undefined }
         | undefined
       const iframe = views?.first?.()?.iframe
--- a/apps/rishi-electron/src/renderer/src/components/epub/EpubView.tsx
+++ b/apps/rishi-electron/src/renderer/src/components/epub/EpubView.tsx
@@ -76,6 +76,7 @@ import { findParagraphForCfi } from '@/modules/cfi-to-paragraph'
 import { resolveLiveSelection } from '@/modules/resolve-live-selection'
 import { buildPartialFirst } from '@/modules/read-aloud-from'
 import { useTtsHighlightReconciler } from '@/hooks/useTtsHighlightReconciler'
+import { useVisibleEpubIframe } from '@/hooks/reader/useVisibleEpubIframe'
 import { createEpubTtsReconciler, type EpubTtsReconciler } from './reconcileTtsHighlight'
 
 function updateTheme(rendition: Rendition, theme: ThemeType) {
@@ -851,31 +852,25 @@ export default function EpubView({ book }: { book: Book }): React.JSX.Element {
   }, [])
 
   const epubTtsReconcilerRef = useRef<EpubTtsReconciler | null>(null)
-  const [epubContentIframe, setEpubContentIframe] = useState<HTMLIFrameElement | null>(null)
+
+  // Track the currently visible content iframe declaratively via
+  // useSyncExternalStore. Previously this was a setState-in-effect cascade
+  // (flagged by react-hooks/set-state-in-effect); the hook now produces the
+  // value during render and re-subscribes only when the rendition identity
+  // changes.
+  const epubContentIframe = useVisibleEpubIframe(rendition)
 
   // Build (or rebuild) the reconciler whenever the rendition instance changes.
+  // Kept in its own effect (separate concern from iframe tracking) — refs are
+  // fine inside effects, and there's no setState here so the React 19 rule
+  // doesn't apply.
   useEffect(() => {
     if (!rendition) {
       epubTtsReconcilerRef.current = null
-      setEpubContentIframe(null)
       return
     }
     epubTtsReconcilerRef.current = createEpubTtsReconciler(rendition)
-
-    // Resolve the epub.js content iframe via the shared helper, which
-    // correctly filters to the currently displayed view. We update the
-    // state on `rendered` so chapter swaps refresh the iframe reference.
-    const resolveIframe = (): void => {
-      const next = getVisibleIframe(rendition) ?? null
-      setEpubContentIframe((prev) => (prev === next ? prev : next))
-    }
-    resolveIframe()
-
-    // epub.js emits 'rendered' when a new view (chapter) is rendered.
-    rendition.on('rendered', resolveIframe)
-
     return () => {
-      rendition.off('rendered', resolveIframe)
       epubTtsReconcilerRef.current = null
     }
   }, [rendition])
--- a/apps/rishi-electron/src/renderer/src/components/epub/reconcileTtsHighlight.ts
+++ b/apps/rishi-electron/src/renderer/src/components/epub/reconcileTtsHighlight.ts
@@ -31,7 +31,7 @@ export function createEpubTtsReconciler(rendition: Rendition) {
     if (desiredIndex && currentTtsCfi !== desiredIndex) {
       const hash = encodeURI(desiredIndex + 'highlight')
       const internal = (rendition.annotations as { _annotations?: Record<string, unknown> })
-        ?._annotations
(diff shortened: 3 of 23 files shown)


matovu-farid (commented):
## Code review funnel — **VERDICT: PASS** (0 blockers)

Reviewed against head SHA \`025f7a9a036b88c325b94bb37bd32e124fafacc8\`.

### Eslint-disable audit ✅
\`git diff origin/main...025f7a9a | grep \"^+\" | grep eslint-disable\` returns nothing. **Zero new disable directives.** The PR's "no shortcuts" commitment is honored end-to-end.

### Real-bug fixes verified ✅
- **\`visualHeuristic.ts:184\`** — \`classifyElement\` (lines 49-88) only ever produces \`kind: 'equation' | 'figure' | null\`. The \`'image'\` arm was provably dead. Standalone \`\` correctly flows through the \`'figure'\` branch with the \`tag === 'img'\` reroute to \`images++\`.
- **\`file-sync.ts:115-141\`** — \`const codeValue = (data as { code?: unknown }).code\`. All three string comparisons run against \`unknown\`; the constructed return uses \`codeValue\` narrowed by the \`===\` checks; \`data.limit\` is still gated with \`typeof === 'number'\`. No bypass path.

### React refactor verdicts ✅
- **\`useVisibleEpubIframe.ts\` (new, 49L)** — Correct \`useSyncExternalStore\` contract. \`subscribe\` attaches \`rendition.on('rendered', onChange)\` and cleans up via \`.off\`. \`getSnapshot\` is null-safe. Both \`useCallback\`s keyed on \`[rendition]\`. Server snapshot \`() => null\` appropriate.
- **\`TTSControls.tsx\` sticky state** — Semantic table matches: \`playing\`→true, \`loading\`→unchanged (guard skips setState), anything else→false. Matches prior semantics; also fixes a first-paint flicker.
- **\`usePdfTextSelection.ts\` + \`pdf.tsx\`** — \`paramsRef\` fully removed, all 5 deps in effect array, 4 call-site \`useCallback\`s with stable empty/ref deps.
- **\`pdf-page.tsx\`** — No surviving \`wrapperRef.current\`. \`wrapper\` state read in render; callbacks include it in deps. Render-gate behavior preserved.

### Other checks
- \`unknown as { manager?: ... }\` casts in 3 files preserve optional chaining so runtime guards remain meaningful — not silencing the rule.
- \`footerMaskByBookId\` consumers all guard against undefined; widening to \`Partial<Record>\` is safe.
- \`Selection.toString().trim()\` per MDN: optional chain handles null Selection, \`toString()\` returns string. Safe.
- \`book.location\` is typed \`string\` at preload boundary — dropping \`?? '0'\` removes a dead fallback (note: separate from #226 which tracks the navigation-history bootstrapping nuance from the \`||\` → \`??\` change in #222).

### CI
Lint in worktree: **0 errors**, 40 warnings — matches the PR's headline claim. Worktree typecheck clean. 1329/1329 tests pass across 146 test files; the 12 \`queries.test.ts\` failures are pre-existing better-sqlite3 ABI mismatches unrelated to this PR.

_See \`.parity/codereview-227/FINDINGS.md\` for the full audit trail._

PR #230, opened 2026-05-23
mobile(reader): inject theme CSS into MOBI/DJVU/PDF WebViews (#47)

### LINKED ISSUE #47: mobile/reader/mobi: MOBI/DJVU/PDF theme changes ignored by WebViews
## Summary
`handleSettingsChange` persists themeName but no `useEffect` forwards theme to the WebView via `injectJavaScript`. Dark-mode users see white pages.

## Files
- `apps/mobile/app/reader/mobi/[id].tsx:494-497`
- `apps/mobile/app/reader/djvu/[id].tsx:509-512`
- `apps/mobile/app/reader/pdf/[id].tsx:553-556`

## Acceptance
- Each reader has a `useEffect` keyed on `settings.themeName` that injects CSS into the WebView matching the active theme.
- Snapshot test toggles theme and verifies `injectJavaScript` is called with theme-specific CSS.

## Domain
`mobile/reader/mobi`, `mobile/reader/djvu`, `mobile/reader/pdf` — fixer batch routing.

## Source
`.parity-v3/2026-05-22-mobile/VALIDATED.md#RDR-033`

matovu-farid:
Closes #47

## Summary

The MOBI, DJVU, and PDF readers persisted `themeName` through `saveReaderSettings` but never forwarded the new theme into the underlying WebView, so dark-mode users were stuck on white pages. Each reader now runs a `useEffect` keyed on `settings.themeName` that calls `injectJavaScript` (or `readerRef.setTheme` for PDF) with a `<style id="rishi-theme">` element built by a shared helper (`lib/reader-theme-css.ts`). The helper is idempotent — repeated calls replace the previous stylesheet rather than stacking — and emits per-theme CSS so toggling between Light/Dark/Sepia yields visibly different output.

For PDF the imperative `PdfWebReaderHandle` gains a typed `setTheme(theme)` so the reader screen can forward CSS through the wrapper without an `as any` cast.

## Testability

TDD

- Red commit: `3c533b4b` — `test(reader): red — theme injection for mobi/djvu/pdf webview (#47)`
- Green commit: `6e7e4f1a` — `fix(reader): green — inject theme CSS into mobi/djvu/pdf webview (#47)`

The new `__tests__/components/reader/webview-theme-injection.test.ts` mixes source-grep (mirrors the existing `mobi-active-paragraph-inject` pattern, since mounting the readers pulls in WebView, expo-file-system, the player store, expo-router) with direct exercise of the shared helper. The helper tests satisfy acceptance bullet 2 explicitly: toggling theme (`white` → `dark`) yields distinct injected JS, and each theme's background/color literal appears in the output.

## Local CI

- typecheck: PASS  (cmd: `pnpm exec tsc --noEmit` — 131 pre-existing errors in `packages/shared/voice-chat`, 0 new errors in changed files)
- test:      PASS  (cmd: `pnpm exec jest __tests__/components/reader/` — 25 suites, 209 tests; full suite has 13 pre-existing FAILs from drizzle-orm/effect/xstate module resolution in packages/shared, unrelated to this change)
- lint:      PASS  (cmd: `pnpm exec eslint <changed-files>` — 0 errors, 3 pre-existing array-type warnings in `PdfWebReader.tsx`)
- format:    SKIP  (no `.prettierrc` in `apps/mobile`; baseline files don't conform to prettier defaults either, so prettier isn't a real gate in this package)
- build:     SKIP  (reason: RN app heavy build)

### DIFF
--- a/apps/mobile/lib/reader-theme-css.ts
+++ b/apps/mobile/lib/reader-theme-css.ts
@@ -1,0 +1,70 @@
+/**
+ * Build the CSS + `injectJavaScript` snippet used by the MOBI / DJVU /
+ * PDF WebView readers to apply a reader theme.
+ *
+ * Issue #47 — the WebView-backed readers (MOBI, DJVU, PDF) used to
+ * persist a theme via `saveReaderSettings` but never forwarded it into
+ * the WebView. Dark-mode users were stuck on the white default. This
+ * helper centralises the CSS so every reader injects an identical
+ * stylesheet keyed by the active `ThemeName`.
+ *
+ * Design notes:
+ *   - We render via a `<style id="rishi-theme">` element so repeated
+ *     injections replace the previous stylesheet instead of stacking.
+ *   - We touch `body` background/color *and* the canvas wrapper used by
+ *     DJVU (`body` is the canvas container) and the pdfjs viewer
+ *     (`#viewerContainer`, `.pdfViewer`) so all three readers visibly
+ *     re-theme on a single helper call.
+ *   - The returned snippet ends in `true;` — react-native-webview
+ *     requires `injectJavaScript` to end in a JS expression, and
+ *     omitting the trailing expression triggers a runtime warning on
+ *     iOS (verified against react-native-webview 13.x).
+ */
+
+import type { ReaderTheme } from '@/types/book'
+
+/**
+ * Return a JavaScript snippet that, when injected via
+ * `WebView.injectJavaScript`, swaps the active reader theme inside the
+ * WebView. The snippet is idempotent — repeated calls replace the
+ * `<style id="rishi-theme">` element rather than appending new ones.
+ */
+export function buildReaderThemeInjection(theme: ReaderTheme): string {
+  const css = buildReaderThemeCss(theme)
+  // JSON.stringify handles escaping (quotes, newlines, backslashes) so
+  // the injected JS literal is always valid no matter what colour
+  // syntax future themes use.
+  const cssLiteral = JSON.stringify(css)
+  return [
+    '(function(){',
+    '  try {',
+    '    var id = "rishi-theme";',
+    '    var existing = document.getElementById(id);',
+    '    if (existing) existing.parentNode.removeChild(existing);',
+    '    var style = document.createElement("style");',
+    '    style.id = id;',
+    `    style.textContent = ${cssLiteral};`,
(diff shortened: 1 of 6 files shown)


apps/mobile/lib/reader-theme-css.ts
```diff
@@ -0,0 +1,70 @@
...
+ * embed it inline in a static HTML template (initial load) while still
+ * using the same colour values as runtime updates.
+ */
+export function buildReaderThemeCss(theme: ReaderTheme): string {
+  const bg = theme.background
+  const fg = theme.color
+  return [
+    `html, body { background: ${bg} !important; color: ${fg} !important; }`,
+    // pdfjs viewer chrome — keep page chrome consistent with body.
+    `#viewerContainer, .pdfViewer { background: ${bg} !important; }`,
```
matovu-farid, line 65: [NIT] The selectors `#viewerContainer, .pdfViewer` are dead code for this PDF reader. The custom pdfjs template at `components/pdf/webview-template.ts` uses `#status` and `#pages` (with `.page-wrapper` children) — not the stock pdfjs IDs (which would only exist if pdfjs's own viewer.html were loaded). The body-background rule above still themes the chrome correctly, so functionally this is fine, but the comment `pdfjs viewer chrome` is misleading. Consider either dropping the rule or replacing it with `#pages` to match the actual template, so a future reader change does not assume this line is doing something.

matovu-farid (commented):
PASS - zero blockers.

The fix is correct end-to-end. Each reader's effect is keyed on `settings.themeName` with a mount gate (`bookLoaded` for MOBI/DJVU, `pageCount > 0` for PDF) so the first injection lands after the WebView template has finished loading. The injected snippet uses a stable `<style id="rishi-theme">` element so toggles replace rather than stack, and ends in `true;` to silence iOS' RN-WebView injectJavaScript warning. PDF correctly routes through the imperative handle (`setTheme`) rather than reaching past `PdfWebReader`. The new test suite covers all three readers plus the helper, and the per-theme background-literal assertion catches the regression where a refactor injects static CSS.

Follow-up worth a separate issue (out of scope here): `components/pdf/webview-template.ts:34` hard-codes `.page-wrapper { background: #fff }`, so in Dark/Sepia the PDF body re-themes but each rendered page keeps a white card around the canvas. Likely intentional (page fidelity), but if Dark mode is supposed to flip the surround too, that is the line `buildReaderThemeCss` would need to override.

PR #235, opened 2026-05-23
fix(electron/tts): debounce play spinner across cached inter-paragraph dips (#232)

### LINKED ISSUE #232: fix(electron/tts): play button shows spinner during cached inter-paragraph transitions
## Summary

The play-button in the TTS pill shows a loading spinner (Loader2, animate-spin) for a brief moment between every paragraph transition, even when the next paragraph's audio is already cached and starts playing instantly. This is a UI regression — audibly the audio is immediate, but visually the spinner flickers between paragraphs, producing unnecessary jitter.

This is the same root cause that commit 082ffac2 already addressed for the **Repeat button** (which used to unmount/remount on every paragraph boundary). The play-button icon was not given the same treatment and still reads `playingState === 'loading'` directly.

## Repro

1. Open a book in Rishi Electron.
2. Start TTS playback.
3. Let it play across multiple paragraph boundaries.
4. Watch the play-button icon at each paragraph boundary.

**Expected:** The play button stays as the pause icon (or whatever the active-playback icon is) when the audio plays through smoothly with no audible gap.

**Actual:** The play button flickers to a spinning Loader2 for a fraction of a second at every paragraph boundary, even though the audio plays instantly because the next paragraph is pre-cached.

## Root cause (confirmed)

- `apps/rishi-electron/src/renderer/src/components/tts/TTSControls.tsx:196-197` renders the spinner whenever `playingState === 'loading'`, regardless of cache hit.
- `packages/shared/src/machines/playerMachine.ts` always transitions to `'loading'` between paragraphs (see e.g. lines 374 / 385 / 432), even when `packages/shared/src/tts/service.ts:50-58` already has a cache hit and emits `audioReady` synchronously.
- Commit 082ffac2 (\"fix(tts): keep Repeat button mounted across inter-paragraph loading\") solved the equivalent jitter for the Repeat button by tracking a `showRepeat` flag that doesn't flip during inter-paragraph dips. The play-button icon needs the same treatment.

## Acceptance criteria

- [ ] When TTS is actively playing and transitions to the next paragraph whose audio is already cached, the play-button icon does **not** flicker to a spinner — it stays on the active-playback (pause) icon.
- [ ] Cold-start playback (first paragraph from a stopped/idle state) still shows the spinner — this case is genuinely loading and the spinner is correct.
- [ ] An inter-paragraph gap that exceeds a reasonable threshold (e.g. cache miss, network stall) still shows the spinner, so users get feedback when something is actually slow.
- [ ] Regression test in `apps/rishi-electron/src/renderer/src/components/tts/TTSControls.test.tsx` proves the play button keeps its active icon across a `playing → loading → playing` transition when the loading dip is brief / the audio is cached. (Mirror the pattern from commit 082ffac2's Repeat test.)
- [ ] No changes to audio playback behavior itself — purely a UI fix.

## Scope

- **In scope:** Electron only (`apps/rishi-electron`). Pure UI change in `TTSControls.tsx`. May add a derived `showSpinner` flag analogous to `showRepeat`, or check the cache via the existing TTS service.
- **Out of scope:** Mobile (separate issue if the same defect appears there — `apps/mobile/components/player/MiniPlayer.tsx` has equivalent `isLoading` rendering). State machine restructure in `packages/shared/`.

matovu-farid:
## Summary
- Suppress the play-button spinner flicker that fired on every paragraph boundary by adding an asymmetric 200ms delay-show debounce to the cached-dip path while keeping cold-start and pause→resume immediate.
- Route both `getPlayIcon()` and the play-button `aria-label` through a single `effectiveState` so screen readers no longer announce "Loading" while the icon stays steady; introduces a "Loading" accessible name on the play button.
- Extends the existing `lastNonLoadingState` latch instead of adding a parallel one. Prev/Next/Stop `disabled` props stay coupled to raw `playingState` because they guard real races.

## Testability
TDD — red-then-green.
- Red commit: `ad59da74` — `test(electron/tts): red — assert play spinner suppressed on cached paragraph transition (#232)`
- Green commit: `b99933b4` — `fix(electron/tts): green — debounce play spinner across cached inter-paragraph dips (#232)`

Six new tests under `TTSControls — play spinner debounce (#232)`:
1. cached `playing → loading → playing` (~50ms) → spinner never appears
2. stalled load (>250ms in `loading`) → spinner appears after debounce
3. cold-start (`idle → loading`) → spinner immediate, no debounce
4. pause→resume (`paused.clean → loading`) → spinner immediate, no debounce
5. hide-immediately when `loading` ends after spinner showed
6. Prev/Next stay disabled while raw `playingState === 'loading'`

## Local CI
```
typecheck: PASS  (cmd: cd apps/rishi-electron && NODE_OPTIONS="--max-old-space-size=8192" npx tsc --noEmit -p tsconfig.web.json --composite false && ... tsconfig.node.json ...)
test:      PASS  (cmd: cd apps/rishi-electron && pnpm test)  1406/1406 passed
lint:      PASS  (cmd: pnpm -F rishi-electron lint)          0 errors (66 pre-existing warnings unrelated to this diff)
format:    PASS  (cmd: prettier --check on the 2 changed files)
build:     SKIP  (heavy electron build; CI handles it)
```

## Scope
Two files only, per the validation funnel's ship conditions:
- `apps/rishi-electron/src/renderer/src/components/tts/TTSControls.tsx`
- `apps/rishi-electron/src/renderer/src/components/tts/TTSControls.test.tsx`

No changes to `playerMachine.ts`, `usePlayerMachine.ts`, `service.ts`, mobile, or shared.

Closes #232

### DIFF
--- a/apps/rishi-electron/src/renderer/src/components/tts/TTSControls.tsx
+++ b/apps/rishi-electron/src/renderer/src/components/tts/TTSControls.tsx
@@ -62,6 +62,14 @@ const ACTIVE_PLAYBACK_STATES: ReadonlySet<PlayerStoreState> = new Set([
   'republishingParagraphs'
 ])
 
+/**
+ * Debounce window (#232) for showing the play-button spinner when the player
+ * dips from an active-playback state into `loading`. Most inter-paragraph
+ * loads with a cache hit clear in <100ms; a 200ms delay-show window
+ * suppresses the flicker without making genuinely slow loads feel laggy.
+ */
+const SPINNER_DEBOUNCE_MS = 200
+
 export default function TTSControls({ bookId, disabled = false }: TTSControlsProps) {
   const prefersReducedMotion = usePrefersReducedMotion()
   const [showError, setShowError] = useState(false)
@@ -192,18 +200,6 @@ export default function TTSControls({ bookId, disabled = false }: TTSControlsPro
     })
   }
 
-  const getPlayIcon = () => {
-    if (playingState === 'loading') {
-      return <Loader2 size={24} className="animate-spin text-black/60" />
-    }
-    if (playingState === 'playing') {
-      return <Pause size={24} className="text-black/60" />
-    }
-    return <Play size={24} className="text-black/60" />
-  }
-
-  const isPlaying = playingState === 'playing'
-
   // Show the Repeat button across a continuous playback session. While
   // `playingState` briefly drops to `loading` between paragraphs, we keep the
   // button mounted so the pill doesn't reflow every paragraph boundary.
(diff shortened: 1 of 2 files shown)


apps/rishi-electron/src/renderer/src/components/tts/TTSControls.tsx
```diff
@@ -222,6 +218,99 @@ export default function TTSControls({ bookId, disabled = false }: TTSControlsPro
...
+      if (spinnerTimerRef.current !== null) {
+        clearTimeout(spinnerTimerRef.current)
+        spinnerTimerRef.current = null
+      }
+    }
+  }, [])
+
+  // `effectiveState` is the play-button's user-facing state: it preserves the
+  // last steady icon during a suppressed loading dip so the icon AND the
+  // aria-label (TTSControls.tsx:359) stay coupled. Prev/Next/Stop `disabled`
```
matovu-farid, line 290: [NIT] Stale line reference: the comment says `TTSControls.tsx:359` but `aria-label={playButtonLabel}` is now at line 448 after this PR added the surrounding state. Future readers grepping by the cited line number will land in the middle of `<style>` JSX. Worth updating to `:448` or just `(see playButtonLabel below)` to avoid drift on the next refactor.

matovu-farid (commented):
**PASS**

Verified all 10 ship conditions: predicate gates on `ACTIVE_PLAYBACK_STATES.has(lastNonLoadingState)` so cold-start (`idle → loading`) and pause→resume (`paused.clean → loading`) show the spinner immediately while cached `playing → loading → playing` dips stay steady. Asymmetric 200ms delay-show / immediate-hide is correctly implemented via `debounceElapsed` + render-set-state-with-guard reset; effect cleanup clears the timer on every re-run and unmount. `effectiveState` drives both the icon AND the button `aria-label` (`playButtonLabel`) together, so the accessible name stays coupled. Prev/Next/Stop `disabled` props continue to read raw `playingState` (lines 436/479/490) — race gates preserved. Loader2 stays decorative (Lucide auto-adds `aria-hidden=true` when no a11y prop is set, confirmed in node_modules/lucide-react/dist/esm/Icon.js:32) so there is exactly one accessible name on the play button.

No-shortcuts greps clean: no eslint-disable / ts-ignore / any / silent catch / skipped tests / setTimeout(0). Scope is exactly the 2 in-bounds files. All 24 tests pass locally including the 6 new debounce specs (cached dip suppressed, stalled load >250ms shows spinner, cold-start immediate, pause→resume immediate, immediate-hide on completion, Prev/Next disabled gating). One stale line-reference comment is the only nit.

matovu-farid (commented):
3-layer code review funnel: **PASS** (auto-merge eligible). 0 surviving findings at Layer 1 (unanimous across 3 reviewers). Layer 2/3 skipped.

**Why clean:** Electron counterpart of #239. Same asymmetric debounce: suppress on active-playback predecessor (cached dip), show immediately on `idle`/`paused.*` predecessor, hide immediately on leaving `loading`. Timer cancellation verified at 3 sites. Prev/Next/Stop disabled gates intentionally still read raw `playingState` (test 6 pins this invariant). Accessibility preserved (label reads `effectiveState`). Auto-collapse test update reflects honest semantic consequence. All no-shortcuts greps clean. 6 new tests exhaustive over predecessor set. `playerMachine.ts` left alone — fix absorbed at render layer.

Posted as COMMENT (not APPROVE) because GitHub blocks self-approval; verdict is in this body. See .parity/codereview-235/FINDINGS.md.

Issue #233, opened 2026-05-23
electron/chat: voice chat never activates in PDF reader (no pdfStore subscription mirroring epubStore)


matovu-farid:
## Summary

Clicking the voice-chat launcher in the **PDF reader** flips the chat UI to the "chatting" state (mic icon → MicOff, `AIChatOrb` appears) but the realtime voice session **never starts** — the AI never connects, the user can talk but is not heard, and nothing is spoken back. The same launcher in the **EPUB reader** works correctly.

## Root cause (verified by file inspection)

`VoiceChatLauncher` only calls `useChatStore.setIsChatting(true)` — it does not call `startChat(bookId)` directly. The actual call to `voice.activate(bookId, …)` that opens the realtime session lives in a **`useChatStore.subscribe`** side-effect inside `epubStore.ts`:

```ts
// src/renderer/src/stores/epubStore.ts:250-261
useChatStore.subscribe(
  (state) => state.isChatting,
  (isChatting) => {
    const bookId = useEpubStore.getState().bookId
    if (isChatting && bookId) {
      useChatStore.getState().startChat(Number(bookId))
    }
  }
)
```

This subscription only fires when `epubStore.bookId` is set — which only happens for EPUBs. **There is no equivalent subscription anywhere in `pdfStore.ts`**, and `VoiceChatLauncher` itself does not bridge the gap. Result: in the PDF reader, `setIsChatting(true)` flips the flag but no listener invokes `voice.activate(book.id, …)`.

Grep evidence (no PDF-side `startChat` call site exists):

```text
$ grep -rn "startChat\\b" src/renderer/src --include="*.ts" --include="*.tsx"
src/renderer/src/stores/chatStore.ts:28   startChat: (bookId: number) => void
src/renderer/src/stores/chatStore.ts:69       startChat: (bookId) => {
src/renderer/src/stores/epubStore.ts:257       useChatStore.getState().startChat(Number(bookId))
src/renderer/src/stores/chatStore.test.ts:…   (tests only)
```

## Repro

1. Open the app, import the PDF fixture `apps/rishi-electron/e2e/fixtures/test-book.pdf` (or any PDF).
2. Open the book; PDF reader mounts. The bottom-right voice-chat launcher (`aria-label="Start voice chat"`) is visible.
3. (For an end-to-end repro you need an authenticated/premium account so the auth dialog does not gate the click. With dev-bypass: `X-Dev-Bypass` header header.)
[...]

PR #241, opened 2026-05-24
fix(electron/chat): activate voice chat in azw3/mobi via shared helper (#237)

### LINKED ISSUE #237: electron/chat: voice chat dead in Azw3View (handles .mobi + .azw3) — no chat-activation subscription
> **Scope refined by `/kanban-validate` funnel (3/3 KEEP at L1+L2+L3, P1, high confidence).** Validation artifacts: `.parity/validate-237/SURVIVORS.json`. The original framing ("MOBI + AZW3 + centralize") was split into two findings — the bug survives, the broad refactor was dropped 3/3 (refactor-as-blocker), but the surviving acceptance bakes in the centralization via an extracted helper.

## Summary

The voice chat launcher renders inside the kindle reader (`Azw3View`, which handles BOTH `.mobi` and `.azw3` per `routes/books.$id.lazy.tsx:88,91-92`) but clicking it never opens a realtime voice session. Premium users on those formats see `setIsChatting(true)` flip the UI flag while `voice.activate()` is never called.

## Root cause

`VoiceChatLauncher.handleClick` only flips `chatStore.isChatting`; the actual `voice.activate(book.id, …)` is triggered by a `useChatStore.subscribe` side-effect. After PR #234 (which fixed the same bug for PDFs), only **two** activation subscriptions exist in the renderer:

- `apps/rishi-electron/src/renderer/src/stores/epubStore.ts:252` — registered inside `initEpubSubscriptions()`; sole caller is `EpubView.tsx:870`.
- `apps/rishi-electron/src/renderer/src/stores/pdfStore.ts:251` — module-scope, gated by `book?.kind !== 'pdf'`.

`Azw3View.tsx` (which serves both `mobi` and `azw3` kinds) calls `useEpubStore.setBookId(book.id)` at line 171 but **never calls `initEpubSubscriptions()`**, so the chat-activation listener is never installed. End-to-end grep of `Azw3View.tsx` for `chatStore`, `useChatStore`, `isChatting`, `startChat`, `voice.activate`, `initEpubSubscriptions` returns zero hits.

> Note: the legacy `MobiView.tsx` is dead code per `routes/books.$id.lazy.tsx:11-13` — both `.mobi` and `.azw3` route through `Azw3View`. **Do not modify `MobiView.tsx`.**

## Mobile parity (Option B shape)

`apps/mobile/components/reader/ReaderOverlay.tsx:60-94` already solves this for the mobile app: a single shared overlay owns the launcher and calls `startChat(bookId, ctx)` directly on tap. No per-format subscription. We should converge electron toward this shape via an extracted helper rather than landing a third copy-paste subscription.

## Acceptance criteria (from validation funnel L3 survivor)

- [ ] **Preferred shape:** extract a tiny shared module `initBookChatSubscription(getBookId)` and call it from `Azw3View`, `EpubView`, and `pdfStore` so the three call sites don't drift again. Closer to mobile parity at `apps/mobile/components/reader/ReaderOverlay.tsx:60-94`. (Senior reviewer will reject a third copy-paste subscription — flagged by L3-V3.)
- [ ] `Azw3View` mounts and unmounts the chat-activation subscription using a **component-local `useRef`** array per the `EpubView.tsx:867-875` pattern. This is required to dodge cross-mount races (rapid format switches).
- [ ] Unit test: render `Azw3View` with a stubbed book, flip `useChatStore.setState({ isChatting: true })`, assert `voice.activate` (or `startChat` spy) called exactly once with the correct `bookId`. Must cover BOTH `book.kind === 'mobi'` and `book.kind === 'azw3'` as separate cases. Also assert cleanup on unmount.
- [ ] E2E parity spec: `apps/rishi-electron/e2e/ai-chat-azw3.spec.ts` mirroring `apps/rishi-electron/e2e/ai-chat-pdf.spec.ts` — open both an `.azw3` book and a `.mobi` book → assert the launcher renders, the premium dialog opens on unauthorized click, and "Maybe later" dismisses. Fixtures `MOBI_FIXTURE` / `AZW3_FIXTURE` already exist in `e2e/helpers/electron-app.ts:9-10`.
- [ ] No regression: `e2e/ai-chat.spec.ts` (EPUB), `e2e/ai-chat-pdf.spec.ts` (PDF), `epubStore.test.ts`, `pdfStore.test.ts` all still pass.
- [ ] Single-fire guarantee: exactly one `startChat` call per `isChatting: false → true` edge per active reader — no double-activation when switching formats.

## Safety analysis (verified by L3-V1)

The L2 alternative "call `initEpubSubscriptions()` wholesale from `Azw3View`" is **safe but uglier** than the extracted helper:

- 4 of 5 unsubs inside `initEpubSubscriptions` are no-ops for Azw3 (gated on `paragraphRendition`, epubjs `Rendition`, `bookOutline` — none of which Azw3View sets).
- `prewarmKey()` is TTL-cached (~9min), de-duped via inflight promise, and does NOT bill audio/transcription tokens — safe to fire on Azw3 book-open.
- `_prefetchTimer` / `_publishPrefetchTimer` early-return on null rendition, so they're no-ops for Azw3.

Use this as a fallback if the extracted helper turns out to be more work than expected — both are correct fixes.

## Out of scope

- The broader Option-A/Option-B refactor proposed in the original issue body (dropped 3/3 at L1 as refactor-as-blocker).
- Migrating PDF away from its inline overlay rendering at `pdf.tsx:1005-1015` to use `<ReaderOverlayControls>`.
- Any change to the voice-chat service itself.

## Affected files

Read:
- `apps/rishi-electron/src/renderer/src/components/azw3/Azw3View.tsx:799` — launcher mount site (no chat wiring)
- `apps/rishi-electron/src/renderer/src/stores/epubStore.ts:143-261` — `initEpubSubscriptions` body (helper extraction source)
- `apps/rishi-electron/src/renderer/src/stores/pdfStore.ts:240-259` — pdfStore subscription (to migrate to shared helper)
- `apps/rishi-electron/src/renderer/src/components/epub/EpubView.tsx:867-875` — `useRef` cleanup pattern to mirror
- `apps/rishi-electron/src/renderer/src/components/chat/VoiceChatLauncher.tsx:27-33` — launcher's setIsChatting call
- `apps/rishi-electron/src/renderer/src/stores/chatStore.ts:55-92` — setIsChatting vs startChat
- `apps/mobile/components/reader/ReaderOverlay.tsx:60-94` — mobile reference implementation

Write:
- `apps/rishi-electron/src/renderer/src/stores/initBookChatSubscription.ts` (new) — shared helper
- `apps/rishi-electron/src/renderer/src/components/azw3/Azw3View.tsx` — call the helper on mount with useRef cleanup
- `apps/rishi-electron/src/renderer/src/stores/epubStore.ts` — replace inline subscription with helper call
- `apps/rishi-electron/src/renderer/src/stores/pdfStore.ts` — replace inline subscription with helper call (keep `book.kind === 'pdf'` guard if helper is generic)
- New unit tests for the helper + Azw3 activation paths
- New e2e spec `e2e/ai-chat-azw3.spec.ts`

PR sizing estimate: ~80-150 LOC total, single ~1-day PR.

matovu-farid:
Closes #237.

## Summary

`Azw3View` (the renderer for both `.mobi` and `.azw3` per `routes/books.$id.lazy.tsx:88,91-92` — `MobiView.tsx` is dead code) mounts `<VoiceChatLauncher />` but never installs the chat-activation subscription that lives only inside `initEpubSubscriptions()`. Result: clicking the launcher on a MOBI/AZW3 book flips `isChatting` and silently does nothing — same bug PDF had in #233.

Fix: extract `initBookChatSubscription(getBookId)` into a tiny shared module and call it from all three reader paths (`Azw3View`, `EpubView` via `initEpubSubscriptions`, and `pdfStore`) so they can't drift again.

## Approach

The L2 alternative — calling `initEpubSubscriptions()` wholesale from `Azw3View` — was rejected at L3 funnel review (the other 4 subs in that function are EPUB-state-gated; coupling `Azw3View` to "epub subscriptions" is a naming/coupling smell). The shared helper is the clean convergence point and brings electron closer to mobile parity at `apps/mobile/components/reader/ReaderOverlay.tsx:60-94`.

- `Azw3View` uses a component-local `useRef` array for cleanup, mirroring `EpubView.tsx:867-875` exactly so rapid format switches don't leak subscriptions.
- `pdfStore`'s `book.kind === 'pdf'` guard is now expressed as `getBookId()` returning `null` for non-PDF books — the helper's contract is "if `getBookId()` returns null, this reader path isn't active right now".
- `EpubView`'s wiring is unchanged at the call site; only the inline `useChatStore.subscribe` block inside `initEpubSubscriptions()` was refactored to delegate to the helper. The other 4 subscriptions in that function are kept intact.

## Testability

TDD path used: 3 red commits → 1 green commit. All red commits were verified to fail for the right reason before implementation.

| | SHA | Purpose |
|---|---|---|
| RED 1 | `d6e0221d` | `test(electron/chat): red — initBookChatSubscription shared helper (#237)` — failing helper unit tests (subscribe/unsubscribe correctness, null-guard, true→false no-fire, getBookId resolution-at-activation-time). |
| RED 2 | `565d6823` | `test(electron/chat): red — azw3 voice chat activation (#237)` — failing Azw3View activation tests covering both `book.kind === 'mobi'` and `'azw3'` plus unmount + remount cleanup. |
| RED 3 | `0ff318a6` | `test(electron/e2e): azw3 + mobi voice chat parity (#237)` — failing e2e parity spec mirroring `ai-chat-pdf.spec.ts`. |
| GREEN | `faad169a` | `fix(electron/chat): activate voice chat in azw3/mobi via shared subscription helper (#237)` — helper + Azw3View wiring + epubStore/pdfStore refactor. |

## Local CI

```
typecheck: PASS (cd apps/rishi-electron && pnpm typecheck)
test:      PASS (cd apps/rishi-electron && pnpm test) — 1412/1412
lint:      PASS (cd apps/rishi-electron && pnpm lint) — 0 errors (66 pre-existing warnings in unrelated files)
format:    PASS (pnpm exec prettier --check <changed files>)
build:     SKIP (electron-builder is heavy; typecheck covers compile)
```

Touched files alone: 0 lint warnings, 0 prettier issues.

EPUB and PDF activation continue to work — verified via the existing `epubStore.test.ts` (18 tests), `pdfStore.test.ts` including the `voice chat activation (#233)` describe block (3 tests covering kind === 'pdf' fires, kind === 'epub' doesn't double-fire, null book doesn't fire), and the existing `chatStore.test.ts` suite all pass in the full vitest run.

## E2E caveat

The e2e parity spec (`ai-chat-azw3.spec.ts`) follows the same shape as `ai-chat-pdf.spec.ts`: launcher renders → click opens premium dialog → "Maybe later" dismisses. The launcher's auth gate fires BEFORE the chat-activation subscription, so the premium dialog appears regardless of the underlying subscription wiring — the real regression guards live in the unit tests added in this PR. The e2e spec runs against both `AZW3_FIXTURE` and `MOBI_FIXTURE` so any future drift between the two kinds in Azw3View is immediately visible.

The new e2e spec was not run locally (heavy electron-launch suite). All non-e2e checks pass clean.

## Validation

Reference: `.parity/validate-237/SURVIVORS.json` — 3/3 KEEP at all funnel layers, P1, high confidence. The acceptance bullets (helper extraction over copy-paste, MOBI + AZW3 unit coverage, e2e parity, no regression for EPUB/PDF) are all addressed above.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm test` clean (1412 tests pass)
- [x] `pnpm lint` 0 errors on touched files
- [x] `pnpm exec prettier --check` clean on touched files
- [ ] Smoke: open a `.mobi` book → click voice-chat launcher → expect premium-auth dialog (without auth) or activation (with auth)
- [ ] Smoke: open a `.azw3` book → same as above
- [ ] Regression: open an `.epub` book → voice chat still works
- [ ] Regression: open a `.pdf` book → voice chat still works (#233)

### DIFF
--- a/apps/rishi-electron/e2e/ai-chat-azw3.spec.ts
+++ b/apps/rishi-electron/e2e/ai-chat-azw3.spec.ts
@@ -1,0 +1,89 @@
+import { test, expect, type Page } from '@playwright/test'
+import {
+  AZW3_FIXTURE,
+  MOBI_FIXTURE,
+  closeApp,
+  closeOverlays,
+  deleteAllBooks,
+  importBook,
+  launchApp,
+  openBook,
+  type LaunchedApp
+} from './helpers/electron-app'
+
+/**
+ * Parity coverage for the AZW3/MOBI reader's voice-chat launcher (#237).
+ *
+ * Same caveat as the PDF version (`ai-chat-pdf.spec.ts`): the launcher's
+ * auth gate fires BEFORE the chat-activation subscription, so the premium
+ * dialog appears on click regardless of the underlying subscription wiring.
+ * This spec therefore mainly guards the observable surface (launcher present,
+ * dialog opens, "Maybe later" dismisses). The real regression guard for the
+ * missing subscription lives in:
+ *   - src/renderer/src/components/azw3/Azw3View.chatActivation.test.tsx
+ *   - src/renderer/src/stores/initBookChatSubscription.test.ts
+ *
+ * Mirrors the EPUB version in `ai-chat.spec.ts` and the PDF version in
+ * `ai-chat-pdf.spec.ts` exactly so any future drift between the four reader
+ * paths (EPUB, PDF, MOBI, AZW3) is immediately visible.
+ *
+ * Both MOBI and AZW3 flow through the same Azw3View renderer
+ * (routes/books.$id.lazy.tsx:88,91-92), so we exercise both kinds against
+ * the same set of assertions to prove the activation wiring is kind-agnostic.
+ */
+
+const fixtures = [
+  { kind: 'azw3' as const, fixture: AZW3_FIXTURE, label: 'AZW3' },
+  { kind: 'mobi' as const, fixture: MOBI_FIXTURE, label: 'MOBI' }
+]
+
+for (const { kind, fixture, label } of fixtures) {
+  test.describe(`Voice chat / premium gate (${label})`, () => {
+    let app: LaunchedApp
+    let bookId: number
+    let bookPage: Page
+
+    test.beforeAll(async () => {
+      app = await launchApp()
+      const book = await importBook(app.page, {
+        fixturePath: fixture,
+        kind,
+        title: `Voice Chat ${label} Spec`
+      })
+      bookId = book.id
+      bookPage = await openBook(app.page, bookId)
+      await expect(bookPage.locator('[aria-label="Start voice chat"]').first()).toBeVisible({
+        timeout: 30000
+      })
+    })
+
+    test.afterAll(async () => {
+      await deleteAllBooks(app.page)
+      await closeApp(app)
+    })
+
+    test.beforeEach(async () => {
+      await closeOverlays(bookPage)
+    })
+
+    test(`voice chat launcher renders in the ${label} reader`, async () => {
+      await expect(bookPage.locator('[aria-label="Start voice chat"]').first()).toBeVisible({
+        timeout: 15000
+      })
+    })
+
+    test(`clicking voice chat without auth opens the premium dialog (${label})`, async () => {
+      await bookPage.locator('[aria-label="Start voice chat"]').first().click()
+      const dialog = bookPage.locator("[data-slot='dialog-content']")
+      await expect(dialog).toBeVisible({ timeout: 5000 })
+      await expect(dialog.locator('text=Sign in').first()).toBeVisible()
+    })
+
+    test(`"Maybe later" closes the premium dialog (${label})`, async () => {
+      await bookPage.locator('[aria-label="Start voice chat"]').first().click()
+      await expect(bookPage.locator("[data-slot='dialog-content']")).toBeVisible({ timeout: 5000 })
+      await bookPage.locator('text=Maybe later').first().click()
+      await expect(bookPage.locator("[data-slot='dialog-content']")).toHaveCount(0)
+    })
+  })
+}
--- a/apps/rishi-electron/src/renderer/src/components/azw3/Azw3View.chatActivation.test.tsx
+++ b/apps/rishi-electron/src/renderer/src/components/azw3/Azw3View.chatActivation.test.tsx
@@ -1,0 +1,140 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { renderHook } from '@testing-library/react'
+import { useEffect, useRef } from 'react'
+
+// Mock the voice-chat service so chatStore.startChat → voice.activate is a
+// spy we can assert on. Mirrors pdfStore.test.ts / chatStore.test.ts setup.
+const { fakeVoice } = vi.hoisted(() => ({
+  fakeVoice: {
+    start: vi.fn(),
+    stop: vi.fn(),
+    activate: vi.fn().mockResolvedValue(undefined),
+    preconnect: vi.fn().mockResolvedValue(undefined),
+    deactivate: vi.fn(),
+    dispose: vi.fn(),
+    prewarmKey: vi.fn(),
+    getState: vi.fn().mockReturnValue('idle' as const),
+    getError: vi.fn().mockReturnValue(null),
+    dismissError: vi.fn(),
+    onStateChange: vi.fn().mockReturnValue(() => {}),
+    onChatStatus: vi.fn().mockReturnValue(() => {}),
+    onEndedByAgent: vi.fn().mockReturnValue(() => {})
+  }
+}))
+
+vi.mock('@/services', () => ({
+  getVoiceChatService: () => fakeVoice
+}))
+
+const { FakeOfflineError } = vi.hoisted(() => ({
+  FakeOfflineError: class OfflineError extends Error {
+    constructor() {
+      super('offline')
+      this.name = 'OfflineError'
+    }
+  }
+}))
+vi.mock('@/services/voice-chat', () => ({ OfflineError: FakeOfflineError }))
+vi.mock('@/utils/sentry', () => ({ captureError: vi.fn() }))
+vi.mock('@/modules/pageCapture', () => ({
+  summarizeCurrentPage: vi.fn().mockReturnValue({ equations: 0, figures: 0, images: 0 })
+}))
+
+import type { Book } from '@/lib/api'
+import { useChatStore } from '@/stores/chatStore'
+import { initBookChatSubscription } from '@/stores/initBookChatSubscription'
+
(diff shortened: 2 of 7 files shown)


matovu-farid (commented):
PASS: no blockers found.

Verified the helper preserves behavior of the two refactored call sites:
- epubStore: old code was `if (isChatting && bookId)` (truthy check on a default-`''` string). New `getBookId` returns `bookId || null`, so empty string folds to null and the helper's nullish-guard bails — semantically identical.
- pdfStore: old guard `book?.kind !== 'pdf'` is preserved by returning `null` from the closure for non-pdf books. Module-scope lifecycle (unsub discarded) matches the previous singleton.
- Azw3View: new mount/unmount lifecycle mirrors EpubView.tsx:867-875 exactly (component-local `unsubsRef`, `useEffect` with cleanup). `key={book.id}` on the route at books.$id.lazy.tsx:92 means the `[book.id]` dep array is effectively a single mount, so no race on book switch.

`Number(bookId)` is safe: `useEpubStore.bookId` is typed `string` (set from `book.id.toString()`) and `Number()` already lived in the pre-diff code; `book.id` is `number` for the other two paths. `isChatting` is typed `boolean`, so `!isChatting` is well-defined.

Unit tests assert observable behavior (fakeVoice.activate spy with correct args + call counts), not zustand internals. No `.skip`/`.only`, no eslint-disable, no @ts-ignore, no empty catches introduced by this diff. The Azw3View test uses a local hook that re-implements the production wiring rather than mounting the real component — acceptable trade-off given the parser/route dependency graph, but it does mean a future drift in Azw3View's actual useEffect body would not be caught by this file alone. Not blocking.

matovu-farid (commented):
3-layer code review funnel: **PASS** (auto-merge eligible). 0 findings raised at Layer 1 (unanimous NO BLOCKERS / NO NITS across all 3 reviewers). Layer 2/3 skipped — nothing to vote on.

**Why clean:** refactor extracts `initBookChatSubscription` helper used by epubStore, pdfStore, and (newly) Azw3View. Behavior preserved at existing call sites; MOBI/AZW3 voice chat now activates correctly (fixes #237). Triple-checked no double-fire path across the three reader paths (mutually exclusive via routing `key={book.id}` + `kind` guards). No-shortcuts greps clean. New unit + e2e tests cover the wiring including unmount/remount cleanup.

Posted as COMMENT (not APPROVE) because GitHub blocks self-approval; verdict is encoded in this body. See .parity/codereview-241/FINDINGS.md for vote breakdown.

Issue #68, opened 2026-05-22
mobile/chat/detail: Source-chip navigation ignores chunk/page location


matovu-farid:
## Summary
`handleSourcePress` discards `source.chunkId`/`source.chapter` and just pushes to the reader root. Citation is non-actionable.

## Files
- `apps/mobile/app/chat/[bookId].tsx:232-241`

## Acceptance
- Pass chunk's CFI / page / chapter as router query param.
- Reader honors `?cfi=...` (epub) or `?page=...` (pdf) on mount.
- Detox test: tap source → reader scrolls to that page.

## Domain
`mobile/chat/detail` — fixer batch routing.

## Source
`.parity-v3/2026-05-22-mobile/VALIDATED.md#CHT-025`

Part of the 2026-05-22 mobile critic sweep.

PR #244, opened 2026-05-24
fix(mobile/chat/detail): clear PulsingDot interval on unmount (#59)

### LINKED ISSUE #59: mobile/chat/detail: PulsingDot interval leaks on unmount
## Summary
The interval created inside the setTimeout callback (line 491) returns its cleanup to the timeout callback, not the useEffect cleanup. Outer cleanup never reaches the `clearInterval`.

## Files
- `apps/mobile/app/chat/[bookId].tsx:475-507`

## Acceptance
- Replace with Reanimated `withRepeat(withSequence(...))` OR hoist the interval ref to a `useRef` and clear it in the outer cleanup.
- Jest test mounts/unmounts the component repeatedly and asserts no leaked intervals.

## Domain
`mobile/chat/detail` — fixer batch routing.

## Source
`.parity-v3/2026-05-22-mobile/VALIDATED.md#CHT-010`

matovu-farid:
Closes #59.

## Summary
PulsingDot's `useEffect` returned its `clearInterval` cleanup from inside the inner `setTimeout` callback, which discards return values. So the outer effect cleanup only ever cleared the timeout — never the interval. After unmount the interval kept firing and called `setOpacity` on a torn-down tree, leaking timers across every appearance of the typing indicator.

Chose the **`useRef`-style hoist** (here a `let` inside the effect closure — same lifetime semantics as a `useRef` for an effect-scoped handle) over a Reanimated `withRepeat(withSequence(...))` rewrite because it is the minimal root-cause fix that keeps the diff inside the PulsingDot area. Issue #68 owns nearby code in the same file, so we deliberately limited scope.

## Implementation
- Hoist the `setInterval` handle to an effect-scoped `let interval` initialized to `null`.
- The deferred `setInterval` (created inside the `setTimeout` callback at `delay` ms) now assigns to that handle.
- The outer cleanup clears the timeout AND `clearInterval(interval)` when it is non-null, then nulls it.
- The `mounted` flag remains so the inner cycle no-ops if either timer fires after unmount but before the cleanup ran (e.g., during the unmount itself).
- Added a named `export function PulsingDot` so the Jest test can render it in isolation. Default export is unchanged.

## Testability
TDD — red commit `9e88fc07`, green commit `ca66e921`.

The test (`apps/mobile/__tests__/chat/pulsing-dot-interval-leak.test.tsx`) mounts `PulsingDot`, lets the deferred `setInterval` fire under fake timers, unmounts, and asserts every id returned by `setInterval` is passed to `clearInterval`. Also runs five mount/unmount cycles to guard against per-cycle leaks. Both cases were red before the fix and green after.

## Local CI
typecheck: PASS (cmd: `node_modules/.bin/tsc --noEmit -p tsconfig.json` — 0 errors in `app/chat/[bookId].tsx` or the new test; pre-existing errors in `packages/shared/src/voice-chat/*`, `__tests__/*` etc. unrelated)
test:      PASS (cmd: `node_modules/.bin/jest __tests__/chat/` — 17 suites / 45 tests)
lint:      PASS (cmd: `pnpm run lint` — 0 errors; 20 pre-existing warnings, none on changed files)
format:    SKIP (no prettier config)
build:     SKIP (heavy)

### DIFF
--- a/apps/mobile/app/chat/[bookId].tsx
+++ b/apps/mobile/app/chat/[bookId].tsx
@@ -697,7 +697,7 @@ function TypingIndicator() {
   )
 }
 
-function PulsingDot({ delay }: { delay: number }) {
+export function PulsingDot({ delay }: { delay: number }) {
   const [opacity, setOpacity] = useState(0.3)
 
   useEffect(() => {
--- a/apps/mobile/app/chat/[bookId].tsx
+++ b/apps/mobile/app/chat/[bookId].tsx
@@ -702,6 +702,12 @@ export function PulsingDot({ delay }: { delay: number }) {
 
   useEffect(() => {
     let mounted = true
+    // #59 — Hoist the interval handle out of the inner `setTimeout`
+    // callback so the outer effect cleanup can actually clear it.
+    // Previously `return () => clearInterval(interval)` was returned
+    // from the timeout callback (which discards return values), leaving
+    // the interval firing forever on torn-down components.
+    let interval: ReturnType<typeof setInterval> | null = null
     const cycle = () => {
       if (!mounted) return
       setOpacity(1)
@@ -713,13 +719,16 @@ export function PulsingDot({ delay }: { delay: number }) {
 
     const timeout = setTimeout(() => {
       cycle()
-      const interval = setInterval(cycle, 600)
-      return () => clearInterval(interval)
+      interval = setInterval(cycle, 600)
     }, delay)
 
     return () => {
       mounted = false
       clearTimeout(timeout)
+      if (interval !== null) {
+        clearInterval(interval)
+        interval = null
+      }
     }
   }, [delay])
 
--- a/apps/mobile/__tests__/chat/pulsing-dot-interval-leak.test.tsx
+++ b/apps/mobile/__tests__/chat/pulsing-dot-interval-leak.test.tsx
@@ -1,0 +1,250 @@
+/**
+ * CHT-010 / Issue #59 — PulsingDot interval leak.
+ *
+ * Bug: the PulsingDot effect (in apps/mobile/app/chat/[bookId].tsx) used to
+ * `return () => clearInterval(interval)` from INSIDE its outer `setTimeout`
+ * callback. That cleanup function was returned to the timeout callback (which
+ * discards return values), NOT to React. Consequently, React's `useEffect`
+ * cleanup only cleared the timeout — never the interval — so the interval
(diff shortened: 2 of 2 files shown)


apps/mobile/app/chat/[bookId].tsx
```diff
@@ -713,13 +719,16 @@ function PulsingDot({ delay }: { delay: number }) {
...
       cycle()
-      const interval = setInterval(cycle, 600)
-      return () => clearInterval(interval)
+      interval = setInterval(cycle, 600)
     }, delay)
 
     return () => {
       mounted = false
       clearTimeout(timeout)
+      if (interval !== null) {
```
Comment (matovu-farid) [inline apps/mobile/app/chat/[bookId].tsx:728]: [NIT] The inner `setTimeout(..., 300)` that resets opacity to 0.3 is not cleared on unmount. It is guarded by the `mounted` flag so it cannot trigger a setState-after-unmount, and it self-clears within 300ms — so this is not a real leak under issue #59's definition. Mentioning only because the cleanup story is now otherwise complete; tracking it would just mean stashing the inner id alongside `interval` and calling `clearTimeout` on it. Up to you whether to tighten this here or punt.

matovu-farid (commented):
PASS: 0 blockers. Fix is correct — hoisting `let interval` to the outer effect closure means the `useEffect` cleanup actually clears the interval (the original bug returned cleanup from inside `setTimeout`, where the return value was discarded). Cleanup is null-safe via the `if (interval !== null)` guard, so an unmount-before-timeout-fires path is safe. Verified by red/green locally: reverting the source on main + adding the `export` causes both tests to fail with `clearedIds === []`; the fixed source passes. Scope is tight — only PulsingDot touched, no drift into `handleSourcePress` (#68 territory). `any` and `eslint-disable` hits in the diff are all in test scaffolding (jest mocks / `require`), which is the established pattern. One minor observation worth checking below.

matovu-farid (commented):
3-layer code review funnel: PASS (auto-merge). 0 nit(s), 0 blocker(s).

L1 (correctness) raised 0 findings — diff is clean across all 3 reviewers. L2/L3 short-circuit (no survivors).

Verified across reviewers: hoist of `let interval` into useEffect outer scope is closure-correct; both unmount paths (before/after delay) handled with null-guard; test spies on setInterval/clearInterval are race-free under single-threaded JS; the inner setTimeout(..., 300) pre-existing and guarded by `mounted` flag (not a regression); export change to PulsingDot benign — all callers verified.

PR #245, opened 2026-05-24
fix(shared/book-import): cleanup bookDir on parse/save failure in picker path (#209)

### LINKED ISSUE #209: mobile/library/import: shared importer leaks bookDir on parse/save failure (picker path)
## Summary
The picker-path importer in `packages/shared/src/book-import/importer.ts:175` does not remove the bookDir on parse-fail or save-fail. PR #207 fixed orphan-dir cleanup in `apps/mobile/lib/file-import.ts` but the shared importer path remains affected. Carved out of #123 because it's outside the original PR's file partition (`packages/shared/`).

## Files
- packages/shared/src/book-import/importer.ts:175 (parse-fail / save-fail branches)

## Acceptance
- [ ] On parse-fail in the picker-path importer, the bookDir is removed before the error is propagated.
- [ ] On save-fail (DB INSERT throws), the bookDir is removed before the error is propagated.
- [ ] Regression tests in `packages/shared/__tests__/` (or equivalent) covering both failure modes.

## Source
Carved out of #123 / PR #207 review (https://github.com/matovu-farid/rishi-monorepo/pull/207#pullrequestreview-4350414906).

matovu-farid:
Closes #209.

## Summary
The picker-path importer in `packages/shared/src/book-import/importer.ts` did not remove the bookDir on parse-fail or save-fail, leaving orphan directories on disk. PR #207 fixed the mobile-side path (copy-throw cleanup in `createMobileFsPort.copyBookToAppData`) but missed the shared package partition — the two later failure surfaces still leaked.

## Implementation
- Added an optional `FsPort.removeBookDir(bookPath)` so platforms owning a per-book directory (mobile) can tear down the whole dir, while electron (flat `userData/<filename>`) can omit it and the importer falls back to `removeFile`.
- Routed both the parse-fail and save-fail branches through a new `rollbackCopy()` helper that prefers `removeBookDir` when defined. The helper swallows cleanup failures so the original parse/save `error` message (and emitted `failed` event) survives unchanged — verified by dedicated regression tests.
- Implemented `removeBookDir` on `createMobileFsPort` (deletes `books/<bookId>/`, using the bookId already baked into the port — no path parsing).

## Testability
TDD — red commit fb6a69aa, green commit 32d2d9ce.

Six new tests in `packages/shared/src/book-import/importer.test.ts`:
- parse-fail prefers `removeBookDir` over `removeFile`
- parse-fail falls back to `removeFile` (electron shape)
- parse-fail preserves the original error when cleanup throws
- save-fail removes the bookDir (new contract — supersedes the old "leaves the copied file on save failure" test)
- save-fail falls back to `removeFile`
- save-fail preserves the original error when cleanup throws

## Local CI
typecheck: PASS  (cmd: `pnpm typecheck` in `packages/shared`)
test:      PASS  (cmd: `pnpm test` in `packages/shared` — 510/510 tests across 36 files)
lint:      PASS  (cmd: `pnpm exec eslint apps/mobile/lib/book-import/adapters.ts` — packages/shared has no eslint config)
format:    SKIP  (no prettier config in repo)
build:     SKIP  (heavy; tsc --noEmit already covers it)

### DIFF
--- a/apps/mobile/lib/book-import/adapters.ts
+++ b/apps/mobile/lib/book-import/adapters.ts
@@ -135,6 +135,22 @@ export function createMobileFsPort(opts: {
         // best-effort
       }
     },
+
+    async removeBookDir(_path: string): Promise<void> {
+      // #209: the importer hands us the path to `book.<ext>` *inside* the
+      // per-book dir. The dir we created in copyBookToAppData is
+      // `<BOOKS_DIR>/<opts.bookId>`. Delete that (and the file inside)
+      // wholesale rather than fishing the bookId out of the path string —
+      // (bookId, format) is already baked into this port.
+      const bookDir = new Directory(BOOKS_DIR, opts.bookId);
+      try {
+        if (bookDir.exists) {
+          bookDir.delete();
+        }
+      } catch {
+        // best-effort — caller has the original parse / save error.
+      }
+    },
   };
 }
 
--- a/packages/shared/src/book-import/importer.test.ts
+++ b/packages/shared/src/book-import/importer.test.ts
@@ -25,9 +25,18 @@ type TestBookId = string;
 export function makeFs(opts?: {
   copyImpl?: (path: string) => Promise<string>;
   removeImpl?: (path: string) => Promise<void>;
-}): { fs: FsPort; removeCalls: string[]; copyCalls: string[] } {
+  removeBookDirImpl?: (path: string) => Promise<void>;
+  /** When true, the FsPort does NOT define `removeBookDir`. */
+  noRemoveBookDir?: boolean;
+}): {
+  fs: FsPort;
+  removeCalls: string[];
+  copyCalls: string[];
+  removeBookDirCalls: string[];
+} {
   const removeCalls: string[] = [];
   const copyCalls: string[] = [];
+  const removeBookDirCalls: string[] = [];
   const fs: FsPort = {
     copyBookToAppData: vi.fn(async (path: string) => {
       copyCalls.push(path);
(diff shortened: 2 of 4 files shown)


matovu-farid (commented):
PASS: 0 blockers. Carve-out from #207 lands cleanly. Verified:

- **Error preservation**: `rollbackCopy` in `packages/shared/src/book-import/importer.ts:70-83` wraps both branches in its own try/catch — the original parse/save `error` string is captured BEFORE the cleanup call, so even if `removeBookDir` throws, the original message propagates unchanged. Two dedicated regression tests (`importer.test.ts:265` and `:336`) exercise this branch explicitly and pass.
- **FsPort surface**: only two implementations exist (`packages/shared/src/book-import/importer.test.ts:31`, `apps/mobile/lib/book-import/adapters.ts:96`). Electron has its OWN `runImport` + `FsIpc` (`apps/rishi-electron/src/renderer/src/services/book-import/types.ts:106`, `.../importer.ts:90`) — totally unrelated code path, so the optional field cannot silently disable cleanup for electron. The PR description's electron-fallthrough framing is a hypothetical that doesn't apply today.
- **Old test wasn't a contract**: 'save failure does NOT roll back copy' was preserving the bug per issue #209's acceptance criteria ("On save-fail, the bookDir is removed before the error is propagated"). Replacing it is correct.
- **Mobile adapter** (`apps/mobile/lib/book-import/adapters.ts:139`): closes over `opts.bookId` from `createMobileFsPort` rather than parsing the path string — safe because the port is constructed per-import at `apps/mobile/lib/book-import/index.ts:49`. No escape-the-root risk; `new Directory(BOOKS_DIR, opts.bookId).delete()` is scoped to the per-book dir.
- **Tests**: 510/510 pass locally. Both parse-fail and save-fail branches have (a) primary cleanup, (b) electron-shape fallback, (c) cleanup-throws preserves error.
- **No-shortcuts grep**: 0 hits (no `as any`, no skipped tests, no empty catches added outside the documented best-effort cleanup pattern that matches the existing `removeFile` precedent at `adapters.ts:115-118`).
- **Scope**: 4 files, all justified.

matovu-farid (commented):
3-layer code review funnel: PASS (auto-merge). 0 nit(s), 0 blocker(s).

Verified across reviewers: rollbackCopy preserves original error (cleanup catch can't overwrite; two regression tests pin); FsPort.removeBookDir is optional + backward-compatible; electron has its own forked importer so doesn't consume shared FsPort; removed test asserted BUGGY behavior — legitimate TDD red→green swap, not test-deletion-to-hide-failures; mobile adapter mirrors existing removeFile best-effort pattern with inline rationale.

PR #247, opened 2026-05-24
fix(electron/reader): re-fire BOOK_OPENED when fresh book CFI arrives (#226)

### LINKED ISSUE #226: EpubView: nav-history dormant for freshly-imported, never-opened books (#222 follow-up)

matovu-farid:
Closes #226.

## Summary
PR #222 swapped `||` → `??` on `EpubView.tsx:112`, but `book.location` is `''` (empty string, not nullish) for freshly-imported books (`main/ipc/formats.ts:129`). The `[book.id]`-keyed lifecycle effect early-returned on `!currentLocation` and did not re-fire when epubjs later seeded a real CFI via `locationChanged`, so `navigationHistoryMachine` stayed in `inactive` for the whole reader session and the back/forward pill was silently disabled until the user closed and reopened the book.

This PR re-fires `BOOK_OPENED` once epubjs writes a real CFI back to `book.location` (around `EpubView.tsx:1078`), so the navigation-history pill works immediately on freshly-imported books — without re-firing on every subsequent page turn (which would clobber the back-stack via `hydrateOnOpen`).

## Implementation
Extracted the EPUB BOOK_OPENED / BOOK_CLOSED lifecycle into a new `useEpubNavHistoryLifecycle` hook so the guard is unit-testable without booting the whole reader. The hook splits the lifecycle into two effects: BOOK_OPENED depends on `[bookId, currentLocation]` and is ref-guarded so it fires at most once per `bookId` (the first time we observe a non-empty CFI); BOOK_CLOSED depends on `[bookId]` and is suppressed when BOOK_OPENED never fired. This removes the previous `eslint-disable react-hooks/exhaustive-deps` — the new dep arrays are honestly exhaustive.

Option A from the issue body, with the obvious naive-version pitfall (per-page-turn `hydrateOnOpen` reset) called out and avoided.

## Testability
TDD — red commit b7d0cee3, green commit 0fc80c84. New test file at `apps/rishi-electron/src/renderer/src/hooks/reader/__tests__/useEpubNavHistoryLifecycle.test.ts` covers: fresh-import no-CFI no-dispatch; first-CFI-arrival exactly one dispatch; subsequent CFI changes no re-dispatch; unmount semantics for both opened and never-opened bookIds; resume case; bookId switch fires close-then-open in order.

## Local CI
typecheck: PASS  (cmd: `pnpm typecheck` — 0 errors)
test:      PASS  (cmd: `pnpm test` — 1425 passed across 158 files)
lint:      PASS  (cmd: `pnpm lint` — 0 errors, 66 pre-existing warnings, none from this PR)
format:    SKIP  (no prettier config in electron)
build:     SKIP  (heavy)

### DIFF
--- a/apps/rishi-electron/src/renderer/src/components/epub/EpubView.tsx
+++ b/apps/rishi-electron/src/renderer/src/components/epub/EpubView.tsx
@@ -86,6 +86,7 @@ import { useBookSyncId } from '@/hooks/reader/useBookSyncId'
 import { useReaderMenuSync } from '@/hooks/reader/useReaderMenuSync'
 import { useCommonMenuHandlers } from '@/hooks/reader/useCommonMenuHandlers'
 import { usePageRequestSubscription } from '@/hooks/reader/usePageRequestSubscription'
+import { useEpubNavHistoryLifecycle } from '@/hooks/reader/useEpubNavHistoryLifecycle'
 import { useSelectionStore } from '@/stores/selectionStore'
 import { useUndoableHighlightShortcut } from '@/hooks/useUndoableHighlightShortcut'
 import { findParagraphForCfi } from '@/modules/cfi-to-paragraph'
@@ -967,17 +968,11 @@ export default function EpubView({ book }: { book: Book }): React.JSX.Element {
 
   const setParagraphRendition = useEpubStore((s) => s.setParagraphRendition)
 
-  // Navigation history: lifecycle (BOOK_OPENED / BOOK_CLOSED)
-  useEffect(() => {
-    if (!currentLocation) return
-    navigationHistoryActor.send({
-      type: 'BOOK_OPENED',
-      bookId: String(book.id),
-      initialPosition: { kind: 'epub', cfi: currentLocation }
-    })
-    return () => navigationHistoryActor.send({ type: 'BOOK_CLOSED' })
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [book.id])
+  // Navigation history: lifecycle (BOOK_OPENED / BOOK_CLOSED).
+  // Extracted to a hook so the ref-guarded re-fire on first-CFI-arrival
+  // (the fix for #226 — freshly-imported books mount with location = '')
+  // can be unit-tested without booting the whole reader.
+  useEpubNavHistoryLifecycle({ bookId: String(book.id), currentLocation })
 
   // Navigation history: PAGE_VISITED on every location change
   useEffect(() => {
--- a/apps/rishi-electron/src/renderer/src/hooks/reader/__tests__/useEpubNavHistoryLifecycle.test.ts
+++ b/apps/rishi-electron/src/renderer/src/hooks/reader/__tests__/useEpubNavHistoryLifecycle.test.ts
@@ -1,0 +1,136 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { renderHook } from '@testing-library/react'
+
+// The lifecycle hook is created in the green-phase commit. Build the specifier
+// from parts so vite's import-analysis can't statically resolve it at transform
+// time — until the hook exists this dynamic import resolves to a
+// module-not-found rejection, which is the expected red-phase failure mode.
+type LifecycleHook = (args: { bookId: string; currentLocation: string }) => void
+
+const HOOK_PATH = '@/hooks/' + 'reader/useEpubNavHistoryLifecycle'
+
+async function loadHook(): Promise<LifecycleHook> {
+  const mod = (await import(/* @vite-ignore */ HOOK_PATH)) as {
+    useEpubNavHistoryLifecycle: LifecycleHook
+  }
+  return mod.useEpubNavHistoryLifecycle
+}
+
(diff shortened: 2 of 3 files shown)


matovu-farid (commented):
PASS: no blockers. Ref-guard correctness verified by walking React effect ordering for fresh-mount, page-turn, unmount, StrictMode double-mount, and in-place bookId-switch cases. BOOK_CLOSED reliably precedes BOOK_OPENED on bookId change because Effect 1 has no cleanup, so Effect 2's cleanup (BOOK_CLOSED) always runs before any setup re-runs. Tests pass (7/7 green at 0fc80c8, all fail with module-not-found at b7d0cee). One worth-checking note: the BOOK_CLOSED-effect closure captures stale bookId via [bookId] deps and the ref matches stale bookId — combined with key={book.id} on EpubView at books.$id.lazy.tsx:91 (so in-place bookId swap can't happen in production) this is fine, but if EpubView ever loses its key, BOOK_OPENED could fire for the new bookId with the prior CFI. Not flagging as a finding because the prerequisite production change doesn't exist.

matovu-farid (commented):
3-layer code review funnel: PASS (auto-merge). 0 nits, 0 blockers.

Verified across reviewers: ref-guard correct under React single-threaded render; '' → cfi transition correctly re-fires BOOK_OPENED once via deps [bookId, currentLocation] + ref short-circuit on page turns; bookId-switch ordering [BOOK_CLOSED, BOOK_OPENED] holds due to React cleanup-then-setup; StrictMode double-invoke produces correct final state (BOOK_OPENED → BOOK_CLOSED → BOOK_OPENED); 7 tests assert event payloads not implementation; previous eslint-disable on exhaustive-deps removed (new deps honest); production has <EpubView key={book.id}> forcing full unmount but hook is defensive anyway.

PR #248, opened 2026-05-24
fix(mobile/reader/pdf): theme .page-wrapper bg in Dark/Sepia (#231)

### LINKED ISSUE #231: mobile/reader/pdf: page-wrapper background stays white in Dark/Sepia themes
## Summary

PR #230 added theme CSS injection for MOBI/DJVU/PDF WebViews so the body background follows the active reader theme (white / dark / sepia). The PDF reader's `webview-template.ts` still hardcodes `.page-wrapper { background: #fff }`, so in Dark or Sepia mode the page surround (the frame around each PDF page) stays bright white — a jarring rectangle against the now-dark/sepia body.

This was flagged as an out-of-scope nit during #230 code review and deferred to its own issue.

## Files

- `apps/mobile/components/pdf/webview-template.ts:33-34` — `.page-wrapper { ...; background: #fff; ... }` hardcoded in the inline `<style>` block.
- `apps/mobile/lib/reader-theme-css.ts` — current theme-injection builder; only touches `html, body`.

## Repro

1. Open any PDF in the mobile reader.
2. Switch theme to Dark.
3. Observe: body background is dark (good), but each page is wrapped in a bright-white margin/box-shadow frame (bad).
4. Same in Sepia.

## Suggested fix

Extend `buildReaderThemeCss(theme)` in `apps/mobile/lib/reader-theme-css.ts` to also include `.page-wrapper { background: <theme.background> }` (or a slightly-elevated surface color from the theme tokens), so the same `<style id=\"rishi-theme\">` element themes the wrapper.

Alternative: keep the wrapper neutral against the body but use a theme-aware surface color (e.g. \`colors.fill.tertiary\` for Dark, \`#f4ecd8\` for Sepia) so it still reads as a 'page' separator without being pure white.

The PDF page CONTENT itself stays white (rendered by pdfjs and that's the source document) — only the surround needs to change.

## Acceptance

- [ ] In Dark theme, `.page-wrapper` background is no longer `#fff`; it blends with the dark body or uses a clearly-themed surface color.
- [ ] In Sepia theme, `.page-wrapper` background uses a sepia-tinted surface color.
- [ ] In Light theme (white), behavior is unchanged.
- [ ] Snapshot or string-match test in `apps/mobile/__tests__/` verifies the injected CSS includes a themed `.page-wrapper` rule for each theme.

## Domain

`mobile/reader/pdf`

## Source

Follow-up from #230 code review (PDF page-wrapper bg out-of-scope nit).

matovu-farid:
Closes #231.

## Summary
PR #230 themed the body background but the PDF reader's `webview-template.ts` hardcoded `.page-wrapper { background: #fff }`, leaving a bright white rectangle around each PDF page in Dark/Sepia modes. This routes the wrapper bg through `buildReaderThemeCss` so it follows the active theme.

## Implementation
Two minimal edits in series:

1. `apps/mobile/lib/reader-theme-css.ts` — `buildReaderThemeCss` now emits an extra `.page-wrapper { background: <theme.background> !important; }` rule. Reusing `theme.background` (rather than introducing a new surface token) keeps the wrapper consistent with the body for every theme — exactly what the issue's acceptance criteria call for (blends with the dark body in Dark mode, sepia-tinted in Sepia, unchanged white in Light).
2. `apps/mobile/components/pdf/webview-template.ts` — the inline `<style>` block no longer hardcodes `background: #fff` on `.page-wrapper`. The injected stylesheet wins via `!important` regardless, but stripping the literal also means the initial load (before the first `injectJavaScript` round-trip) inherits the body colour instead of flashing white. pdfjs renders the actual page CONTENT to a canvas, so the wrapper colour only shows in the gutters and before-render moments — exactly the surface that needs to follow the theme.

## Testability
TDD — red commit 1c3b88d3, green commit 2993a007.

New `__tests__/components/reader/pdf-page-wrapper-theme.test.ts` covers:
- For each `ThemeName` (`white`/`dark`/`yellow`), the CSS produced by `buildReaderThemeCss` contains a `.page-wrapper` rule whose declaration block embeds the theme's `background` literal.
- The dark stylesheet must not contain `#fff` / `#ffffff` (regression pin against re-introducing the hardcoded literal).
- The PDF webview template source no longer contains a `.page-wrapper { ... background: #fff ... }` declaration block (whitespace-insensitive match).

All existing tests in `webview-theme-injection.test.ts` (PR #230) still pass.

## Local CI
```
typecheck: PASS  (cmd: ./node_modules/.bin/tsc --noEmit -p tsconfig.json) — 0 errors in changed files; pre-existing errors in packages/shared/voice-chat & unrelated test files predate this branch
test:      PASS  (cmd: ./node_modules/.bin/jest __tests__/components/reader/ __tests__/pdf/pdf-webview-bridge.test.ts) — 28 suites, 243 tests
lint:      PASS  (cmd: pnpm run lint) — 0 errors, 20 pre-existing warnings (none in changed files)
format:    SKIP  (no prettier config)
build:     SKIP  (heavy)
```

### DIFF
--- a/apps/mobile/components/pdf/webview-template.ts
+++ b/apps/mobile/components/pdf/webview-template.ts
@@ -30,8 +30,13 @@ export const PDF_READER_HTML = `<!DOCTYPE html>
     -webkit-user-select: text; -webkit-touch-callout: default; }
   #status { padding: 12px; font-size: 13px; }
   #pages { padding: 8px 0 64px 0; }
+  /* Issue #231 — wrapper background is themed at runtime via
+     buildReaderThemeCss / buildReaderThemeInjection. Leaving a static
+     #fff here reintroduces the bright rectangle in Dark/Sepia mode
+     because the inline rule would shadow the injected one on the
+     initial load. */
   .page-wrapper { position: relative; margin: 8px auto;
-    background: #fff; box-shadow: 0 1px 6px rgba(0,0,0,0.6); }
+    box-shadow: 0 1px 6px rgba(0,0,0,0.6); }
   .page-canvas { display: block; width: 100%; height: auto; }
   .text-layer { position: absolute; inset: 0; line-height: 1;
     overflow: hidden; pointer-events: auto; }
--- a/apps/mobile/lib/reader-theme-css.ts
+++ b/apps/mobile/lib/reader-theme-css.ts
@@ -63,6 +63,16 @@ export function buildReaderThemeCss(theme: ReaderTheme): string {
     `html, body { background: ${bg} !important; color: ${fg} !important; }`,
     // pdfjs viewer chrome — keep page chrome consistent with body.
     `#viewerContainer, .pdfViewer { background: ${bg} !important; }`,
+    // Issue #231 — the custom PDF WebView template (components/pdf/
+    // webview-template.ts) renders each pdfjs page inside a
+    // `.page-wrapper` div. Its inline stylesheet previously hardcoded
+    // `background: #fff`, leaving a bright rectangle around every page
+    // in Dark / Sepia mode. Routing the wrapper background through the
+    // active theme keeps the page surround consistent with the body.
+    // pdfjs paints the actual page content onto a canvas, so the
+    // wrapper colour only shows in the gutters and before-render
+    // moments — exactly the surface that needs to follow the theme.
+    `.page-wrapper { background: ${bg} !important; }`,
     // MOBI parser content wrapper.
     `#content, #content * { color: ${fg} !important; }`,
     `#content a { color: ${fg} !important; }`,
--- a/apps/mobile/__tests__/components/reader/pdf-page-wrapper-theme.test.ts
+++ b/apps/mobile/__tests__/components/reader/pdf-page-wrapper-theme.test.ts
@@ -1,0 +1,98 @@
+/**
+ * Issue #231 — PR #230 (issue #47) themed the WebView body background but
+ * the PDF reader's inline `<style>` block in `webview-template.ts`
+ * hardcodes `.page-wrapper { background: #fff }`. In Dark / Sepia themes
+ * the body re-themes, but the white rectangle behind each PDF page stays
+ * bright white — a jarring contrast against the dark surround.
+ *
+ * Acceptance:
+ *   - Each ReaderTheme yields injected CSS that includes a `.page-wrapper`
+ *     rule whose background matches the active theme's `background`.
+ *   - Light theme behaviour stays visually identical (still resolves to
+ *     the theme's `#FFFFFF`).
+ *   - The PDF webview template no longer hardcodes a static
+ *     `background: #fff` on `.page-wrapper` — the injected CSS must win.
+ *
+ * Source-grep is the established pattern for the PDF WebView template
+ * (see `pdf-webview-bridge.test.ts`) — instantiating the HTML/JS inside
+ * Jest would require a real DOM + pdfjs runtime. Pinning the hardcoded
+ * literal at the source level is the cheapest way to catch a regression
+ * that re-introduces the white rectangle.
+ */
+
+import { readFileSync } from 'fs'
+import { join } from 'path'
+
+import { READER_THEMES } from '../../../constants/reader-themes'
+import type { ThemeName } from '../../../types/book'
+
+const MOBILE_ROOT = join(__dirname, '..', '..', '..')
+
+const THEME_NAMES: ThemeName[] = ['white', 'dark', 'yellow']
+
+describe('Issue #231 — PDF .page-wrapper background follows the active theme', () => {
+  describe('shared reader-theme CSS helper', () => {
+    let buildReaderThemeCss: (theme: (typeof READER_THEMES)[ThemeName]) => string
+
+    beforeAll(async () => {
+      const mod = await import('../../../lib/reader-theme-css')
+      buildReaderThemeCss = mod.buildReaderThemeCss
+    })
+
+    it.each(THEME_NAMES)(
+      'emits a .page-wrapper rule using the %s theme background',
+      (name) => {
+        const theme = READER_THEMES[name]
+        const css = buildReaderThemeCss(theme)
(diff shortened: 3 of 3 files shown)


apps/mobile/components/pdf/webview-template.ts
```diff
@@ -30,8 +30,13 @@ export const PDF_READER_HTML = `<!DOCTYPE html>
     -webkit-user-select: text; -webkit-touch-callout: default; }
   #status { padding: 12px; font-size: 13px; }
   #pages { padding: 8px 0 64px 0; }
+  /* Issue #231 — wrapper background is themed at runtime via
+     buildReaderThemeCss / buildReaderThemeInjection. Leaving a static
+     #fff here reintroduces the bright rectangle in Dark/Sepia mode
+     because the inline rule would shadow the injected one on the
+     initial load. */
   .page-wrapper { position: relative; margin: 8px auto;
```
matovu-farid, line 38: [NIT] First-paint behavior note (not a blocker, no action required): the inline `<style>` still sets `html, body { background: #1a1a1a }` on line 28. With the static `.page-wrapper` background gone, wrappers now inherit that dark gray until `buildReaderThemeInjection` lands. For Dark theme that is a strict improvement; for Light/Sepia users there is a brief dark-gray flash on the wrapper rectangles before the injected stylesheet repaints them — whereas previously the inline `#fff` matched the final Light state exactly. The net trade is correct (Dark was visibly broken, Light flash is sub-frame in practice), but if you want zero flash across all themes the body bg on line 28 would also need to be parametrized at template build time (not the injection — the static HTML literal). Flagging so it is a deliberate choice rather than an oversight.   -f 'comments[][path]=apps/mobile/lib/reader-theme-css.ts'   -F 'comments[][line]=75'   -F 'comments[][side]=RIGHT'   -F 'comments[][body]=[NIT]  here is no longer load-bearing — the inline  rule in  no longer declares a  at all, so a non- rule would already win on specificity. Keeping  is consistent with the other rules in this file (defensive against future inline rules or pdfjs internal CSS), which is the right call. Just noting it is consistency-driven, not specificity-driven.'

matovu-farid (commented):
PASS — no blockers. Fix is correctly scoped: removing the inline  lets the injected  rule (theme.background) win on both initial and runtime injection. Verified: (a)  so Light is visually identical; (b)  is only used by the PDF webview template — MOBI/DJVU/EPUB readers have no parallel exposure; (c) no  /  / . One worth-noting non-blocker on first-paint flash — see inline.

apps/mobile/components/pdf/webview-template.ts
matovu-farid, line 38: [NIT] First-paint behavior note (not a blocker, no action required): the inline `<style>` on line 28 still sets `html, body { background: #1a1a1a }`. With the static `.page-wrapper` background gone, wrappers now inherit that dark gray until `buildReaderThemeInjection` lands.

For Dark theme this is a strict improvement. For Light/Sepia users there is a brief dark-gray flash on the wrapper rectangles before the injected stylesheet repaints them — whereas the old inline `#fff` matched the final Light state exactly.

Net trade is correct (Dark was visibly broken, Light flash is sub-frame in practice), but if you want zero flash across all themes the body bg on line 28 would also need to be parametrized at template build time (in the static HTML literal, not via injection). Flagging so it is a deliberate choice rather than an oversight.

apps/mobile/lib/reader-theme-css.ts
```diff
@@ -63,6 +63,16 @@ export function buildReaderThemeCss(theme: ReaderTheme): string {
...
+    // Issue #231 — the custom PDF WebView template (components/pdf/
+    // webview-template.ts) renders each pdfjs page inside a
+    // `.page-wrapper` div. Its inline stylesheet previously hardcoded
+    // `background: #fff`, leaving a bright rectangle around every page
+    // in Dark / Sepia mode. Routing the wrapper background through the
+    // active theme keeps the page surround consistent with the body.
+    // pdfjs paints the actual page content onto a canvas, so the
+    // wrapper colour only shows in the gutters and before-render
+    // moments — exactly the surface that needs to follow the theme.
+    `.page-wrapper { background: ${bg} !important; }`,
```
matovu-farid, line 75: [NIT] `!important` here is no longer load-bearing — the inline `.page-wrapper` rule in `webview-template.ts` no longer declares a `background` at all, so a non-`!important` rule would already win on specificity (and the injected `<style>` appended to `document.head` comes later in source order anyway). Keeping `!important` is consistent with the other rules in this helper (defensive against future inline rules or pdfjs internal CSS), which is the right call. Just noting that it is consistency-driven, not specificity-driven, so a future reviewer does not look for the conflicting rule that justifies it.

matovu-farid (commented):
PASS — no blockers. Fix is correctly scoped: removing the inline `background: #fff` lets the injected `.page-wrapper` rule (theme.background) win on both initial and runtime injection.

Verified:
- `READER_THEMES.white.background === '#FFFFFF'` so Light is visually identical to the old hardcoded `#fff`.
- `.page-wrapper` is only used by the PDF webview template — MOBI/DJVU/EPUB readers have no parallel exposure (greped `apps/mobile/components/`).
- No `as any` / `@ts-ignore` / `.skip` / `.todo` introduced.
- Test asserts `.page-wrapper\b` selector presence + theme.background substring + negative `#fff` / `#ffffff` in the dark stylesheet — pins behavior without snapshot brittleness.

Two non-blocker notes inline.

matovu-farid (commented):
3-layer code review funnel: PASS (auto-merge). 0 nits, 0 blockers.

Verified across reviewers: theme.white.background='#FFFFFF' (Light visually identical to old hardcoded #fff); !important consistent with sibling rules (5/5 in helper carry it) — not specificity-required after hardcoded rule removal but matches house style; no first-paint race (pre-existing body bg from inline <style> handles); grep confirms .page-wrapper ONLY in PDF webview-template (MOBI/DJVU don't have parallel exposure); test regex assertions correct (whitespace-stripped template grep, per-theme bg substring, #fff/#ffffff regression pin).

PR #214, opened 2026-05-23
fix(electron/sync): validate IPC payloads + wrap multi-statement ops in tx

matovu-farid:
Closes #166. Closes #167.

## Summary
- Validate sync IPC payloads (`applyBookConflict`, `applyHighlightConflict`, `applyConversationConflict`, `upsertBook`) with shared zod schemas in `sync.schemas.ts` instead of casting `Record<string, unknown>` field by field. Malformed cloud payloads coerce to safe defaults or fail fast — no garbage in SQLite.
- Wrap multi-statement sync ops in `db.transaction(...)` so a partial failure rolls back.
- Extract the four conflict handlers + mark-clean loops into `_*WithDb` helpers that take an explicit Drizzle instance, enabling true unit tests.

## Test plan
- [x] `pnpm test src/main/ipc/__tests__/sync-validation.test.ts` — 9/9 pass
- [x] `pnpm typecheck` — clean

PR #253, opened 2026-06-01
feat(sharing): shared-reading sessions over Cloudflare Worker + WebRTC mesh

matovu-farid:
## Summary
- **Worker + DO** (`workers/sharing-worker/`): Cloudflare Worker with Durable Object signaling, WS hibernation, Better Auth verification, HMAC-signed join/reconnect tokens, base64url subprotocol bearer, `sync.frame` + `data.channel.relay` server-side relays. Deployed to `https://rishi-sharing-worker.faridmato90.workers.dev` (v `1329671a`). 75/75 worker tests.
- **Electron client** (`apps/rishi-electron/`): XState v5 sessionMachine with `signalingActor` + `peerActor` + `syncActor` + `reconnectActor` + `fileTransferActor`, IPC + main-process auth/library/deep-link wiring, full UI surface (SessionPanel, HostSuspendedBanner, KickedDialog, RoleTransferToast, SessionEndedKeepBooksDialog, ApprovalWaitingScreen, etc.) mounted via `SharingSessionOverlay`. 1750+ unit tests.
- **E2E suite** (Playwright + 2 Electron processes + wrangler dev): 9/10 sharing tests passing, 1 \`test.fixme\` (file-transfer needs product-side fileTransferActor invoke wire-up).
- **Rollout machinery**: \`isSharingEnabled()\` + \`isSharingEnabledForUser()\` shard gate, Sentry \`feature=sharing\` tag + 8 breadcrumbs, rollout runbook with telemetry thresholds.

## What's behind the flag
The feature is gated by \`VITE_SHARING_ENABLED\` (build-time) + \`localStorage['rishi:sharing-enabled']=1\` (runtime opt-in). Internal testers can opt in via DevTools console; broader rollout uses \`VITE_SHARING_ROLLOUT_PCT\` with deterministic per-user bucket.

## Known gaps (tracked, not blocking)
- File-transfer E2E (\`test.fixme\`): viewer-without-book path needs sessionMachine to invoke fileTransferActor on \`hello.hasBookFile=false\`. Relay infrastructure (\`data.channel.relay\`) is ready.
- Reader-toolbar entry button only visible to authenticated users in the rollout bucket (intentional).
- Logpush observability skipped; Sentry covers errors. Decision documented in \`workers/sharing-worker/logpush/saved-queries.md\` (config files committed but no R2 bucket / job).

## Test plan
- [ ] CI is green (worker tests, renderer unit tests, lint, typecheck)
- [ ] Manual smoke in dev: \`pnpm dev\`, DevTools: \`localStorage.setItem('rishi:sharing-enabled','1') + location.reload()\`, open a book, click Share, copy invite, paste into a second instance, confirm both reach live with SessionPanel visible.
- [ ] E2E suite locally: \`cd apps/rishi-electron && pnpm test:e2e:sharing\` — expect 9 passed, 1 fixme.
- [ ] Verify production worker health: \`curl https://rishi-sharing-worker.faridmato90.workers.dev/health\` → \`ok\`.


matovu-farid:
### Code review

Found 3 issues:

1. Durable Object's in-memory `pendingSockets` Map (alongside `lastRequestSharer`, `frameBuckets`) is not repopulated on hibernation wake. WebSocket Hibernation is in use (`ctx.acceptWebSocket` at line 174), so the JS instance can be evicted while `state.pendingJoiners` persists in storage. On wake, `approve.join` reads `pendingSockets.get(msg.userId)` → undefined, silently deletes the pending entry, and `break`s — the joining user is never admitted and the host gets no error.

https://github.com/matovu-farid/rishi-monorepo/blob/bd02c2dbe41de4071c600924f3169c4be1668228/workers/sharing-worker/src/SessionRoom.ts#L308-L322

2. `sharing:*` IPC handlers pass `params` straight through with no runtime validation. `sharing:discardTransferredBook` calls `fs.unlink(params.localPath)` — a renderer with XSS (or a future code path that surfaces peer-influenced paths) can unlink arbitrary files. `sharing:saveTransferredBook` accepts `format` typed as `'epub' | 'pdf'` but unvalidated at runtime, so a bad value flows into the on-disk path. Past PR #214 introduced zod schemas on `sync:*` IPC for exactly this trust-boundary reason; sharing IPC has the same threat model since book bytes arrive via WebRTC.

https://github.com/matovu-farid/rishi-monorepo/blob/bd02c2dbe41de4071c600924f3169c4be1668228/apps/rishi-electron/src/main/ipc/sharing.ts#L26-L43

3. `persistAndReportReceivedBook` fires `saveTransferredBook` as `void`-then-.catch with only a `console.warn` on failure. The receiver actor has already emitted `TRANSFER_RECEIVED` and gets stopped synchronously in the same transition, so on failure (disk full, perms, db lock) the user has no UI signal and no retry path — silent data loss. The .catch should at least send a failure event to the machine and surface a toast.

https://github.com/matovu-farid/rishi-monorepo/blob/bd02c2dbe41de4071c600924f3169c4be1668228/apps/rishi-electron/src/renderer/src/machines/sessionMachine.ts#L687-L718

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

